PR #193, opened 2013-10-23
Implement the `preparer` and `validator` run, for the root SchemaNode.

abourget:
Colander runs them, but Cornice did not.

Tests run.  Haven't tested with py3k though?!


leplatrem:
I believe we can close that in favor of #330. Thoughts ?

abourget:
Close close.

Le lun. 14 sept. 2015 06:00, Mathieu Leplatre <notifications@github.com> a
écrit :

leplatrem:
All right, thanks for your reply!

PR #261, opened 2014-11-04
Set cors headers when a response is returned or exception is raised

circlingthesun:
When a view raises an exception or returns a Response object, cors allow-origin headers are not set. This PR makes sure the headers are set.


Natim:
@circlingthesun any news?

circlingthesun:
@Natim, @ametaireau tests were just added.

Natim:
That's great. Thanks.

ametaireau:
Thanks!

PR #256, opened 2015-01-28
Empty response to OPTIONS request

Natim:
While using angular to make POST request I've noticed a warning:

> 2014-10-12 21:28:12,091 WARNI [cornice][Dummy-2] returning a json string or array is a potential security hole, please ensure you really want to do this.

After some investigation I realised that this is because response to OPTIONS request had an empty body `b'""'`. If I'm not mistaken, angular makes this request before POSTing data. Would it be better if cornice(?) returned an empty object `{}`?

To replicate this warning define any resource. Add `cors_origins=('*',)` option, and then make a call, for example, with `curl`:

> curl -X OPTIONS -H "Origin:http://localhost:6543/" -H "Access-Control-Request-Method:OPTIONS" localhost:6543/api/v1/notes

And you'll get a warning from cornice.

### DIFF
--- a/cornice/tests/test_validation.py
+++ b/cornice/tests/test_validation.py
@@ -10,7 +10,6 @@ from pyramid.response import Response
 from cornice.errors import Errors
 from cornice.tests.validationapp import main, includeme, dummy_deserializer
 from cornice.tests.support import LoggingCatcher, TestCase, CatchErrors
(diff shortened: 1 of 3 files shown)


ametaireau:
Yeah, I believe that's the better approach. This isn't really a problem and is causing more trouble that it actually solves.

PR #271, opened 2015-02-05
Ensure passed JSON is an object.

ametaireau:
Otherwise, the colander validation on body (if any) may fail and throw a 500.

@leplatrem r?

### DIFF
--- a/cornice/util.py
+++ b/cornice/util.py
@@ -128,7 +128,13 @@ def extract_json_data(request):
     if request.body:
         try:
             body = simplejson.loads(request.body)
-            return body
+            if isinstance(body, dict):
+                return body
+            request.errors.add(
+                'body', None,
+                "Invalid JSON: Should be a JSON object, got %s" % body
+            )
+            return {}
         except ValueError as e:
             request.errors.add(
                 'body', None,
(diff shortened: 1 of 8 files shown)


cornice/util.py
```diff
@@ -128,7 +128,13 @@ def extract_json_data(request):
     if request.body:
         try:
             body = simplejson.loads(request.body)
-            return body
+            if isinstance(body, dict):
```
leplatrem: What if the body is an array ?

Natim: Valid JSON can be string, array or dict. But what would be the purpose of sending such a body to cornice? Colander validators works on object only, isn't it?

leplatrem: No colander schemas can be sequences!

```
PUT /collection
[
  {"record": 1},
  {"record": 2},
  {"record": 3}
]

ametaireau: It seems that cornice only supports mapping in the way it's implemented currently.

leplatrem: Then maybe we could add a safety check in Service asserting that the given colander schema inherits from colander.MappingSchema ?

ametaireau: yeah, that's the direction I think we should take as well!

Natim: Actually schema does inherit from ``CorniceSchema`` I added an assertion to make sure it works.

PR #272, opened 2015-02-05
Plug travis with tox

ametaireau:
It seems better to use tox for that.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 python: 2.7
 sudo: false
-services: redis-server
 env:
     - TOX_ENV=py26
     - TOX_ENV=py27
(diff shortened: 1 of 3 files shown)


ametaireau:
@Natim @leplatrem r?

.travis.yml
```diff
@@ -1,10 +1,17 @@
 language: python
-python:
-    - "2.6"
-    - "2.7"
-    - "3.3"
-install: pip install --use-mirrors .
-script: python setup.py test
+python: 2.7
+sudo: false
+services: redis-server
```
Natim: Do you really need that?

ametaireau: nope !)

ametaireau:
updated

PR #273, opened 2015-02-12
When using set_default_permissions OPTIONS calls for CORS does not work anymore.

Natim:
Because browser doesn't send Authorization headers on OPTIONS call.

### DIFF
--- a/cornice/pyramidhook.py
+++ b/cornice/pyramidhook.py
@@ -159,7 +159,8 @@ def register_service_views(config, service):
     # before doing anything else, register a view for the OPTIONS method
     # if we need to
     if service.cors_enabled and 'OPTIONS' not in service.defined_methods:
-        service.add_view('options', view=get_cors_preflight_view(service))
+        service.add_view('options', view=get_cors_preflight_view(service),
+                         permission=NO_PERMISSION_REQUIRED)
 
     # register the fallback view, which takes care of returning good error
     # messages to the user-agent
--- a/cornice/tests/test_cors.py
+++ b/cornice/tests/test_cors.py
@@ -62,7 +62,7 @@ def gimme_some_spam_please(request):
     return 'spam'
 
 
-@spam.post(permission="read-only")
+@spam.post(permission='read-only')
 def moar_spam(request):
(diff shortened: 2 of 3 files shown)


cornice/pyramidhook.py
```diff
@@ -159,7 +159,8 @@ def register_service_views(config, service):
     # before doing anything else, register a view for the OPTIONS method
     # if we need to
     if service.cors_enabled and 'OPTIONS' not in service.defined_methods:
-        service.add_view('options', view=get_cors_preflight_view(service))
+        service.add_view('options', view=get_cors_preflight_view(service),
+                         permission=NO_PERMISSION_REQUIRED)
```
ametaireau: Good catch thanks!

ametaireau:
Can you add a test to reflect that?

Natim:
I tried to reproduce using a test but I wasn't able to with service. I will try with Resource.

Natim:
Here we go :)

cornice/tests/test_cors.py
```diff
@@ -98,12 +101,11 @@ def setUp(self):
         self.config = testing.setUp()
         self.config.include("cornice")
         self.config.add_route('noservice', '/noservice')
-
         self.config.scan("cornice.tests.test_cors")
         self.app = TestApp(CatchErrors(self.config.make_wsgi_app()))
 
-        def tearDown(self):
-            testing.tearDown()
+    def tearDown(self):
```
Natim: oops

cornice/tests/test_cors.py
```diff
@@ -312,3 +314,36 @@ def test_existing_non_service_route(self):
         resp = self.app.get('/noservice', status=200,
                             headers={'Origin': 'notmyidea.org'})
         self.assertEqual(resp.body, b'No Service here.')
+
+
+class TestAuthenticatedCors(TestCase):
```
Natim: CORS, upper case

Natim:
(@ametaireau speaking) that works for me. Changes to do: single to double quotes everywhere, update Cors in CORS and rename "noservice" in something else.

PR #278, opened 2015-02-26
Wrong if clause that passed even if wrong on Python 2.*

Natim:
r? @ametaireau

### DIFF
--- a/cornice/cors.py
+++ b/cornice/cors.py
@@ -125,7 +125,7 @@ def apply_cors_post_request(service, request, response):
             'Access-Control-Allow-Credentials' not in response.headers):
         response.headers['Access-Control-Allow-Credentials'] = 'true'
 
-    if request.method is not 'OPTIONS':
+    if request.method != 'OPTIONS':
         # Which headers are exposed?
         supported_headers = service.cors_supported_headers
         if supported_headers:
--- a/cornice/tests/test_cors.py
+++ b/cornice/tests/test_cors.py
@@ -139,6 +139,14 @@ class TestCORS(TestCase):
             status=400)
         self.assertEqual(len(resp.json['errors']), 1)
 
+    def test_preflight_does_not_expose_headers(self):
+        resp = self.app.options(
+            '/squirel',
+            headers={'Access-Control-Request-Method': 'GET',
+                     'Origin': 'notmyidea.org'},
+            status=200)
+        self.assertNotIn('Access-Control-Expose-Headers', resp.headers)
+
     def test_preflight_missing_request_method(self):
 
         resp = self.app.options(


cornice/cors.py
```diff
@@ -125,7 +125,7 @@ def apply_cors_post_request(service, request, response):
             'Access-Control-Allow-Credentials' not in response.headers):
         response.headers['Access-Control-Allow-Credentials'] = 'true'
 
-    if request.method is not 'OPTIONS':
+    if request.method != 'OPTIONS':
```
ametaireau: I cannot reproduce with my shell, on python 2.6.9 and python 2.7.6.

Do you have more information about that?

```python
>>> method = "OPTIONS"
>>> method is not "OPTIONS"
False
>>> method = "NOT OPTIONS"
>>> method is not "OPTIONS"
True
```

Natim: Just turn it back and run the test.

ametaireau: On python 2.7.5+, I don't have the problem when turning it back and running the test, with a `pdb.set_trace()`:

```
> /home/alexis/dev/github.com/cornice/cornice/cors.py(129)apply_cors_post_request()
-> if request.method != 'OPTIONS':
(Pdb) request.method != 'OPTIONS'
False
(Pdb) request.method is not 'OPTIONS'
False
```

ametaireau: Okay, I just reproduced, this is weird because not reproductible each time.
``` .tox/py27/bin/nosetests -s cornice.tests.test_cors:TestCORS.test_preflight_does_not_expose_headers
> /home/alexis/dev/github.com/cornice/cornice/cors.py(129)apply_cors_post_request()
-> if request.method != 'OPTIONS':
(Pdb) request.method
'OPTIONS'
(Pdb) request.method is not 'OPTIONS'
True
(Pdb) request.method != 'OPTIONS'
False
```

ametaireau: From inside tox:

```python
(Pdb) id(request.method)
38864240
(Pdb) id('OPTIONS')
139946937692256
```

Directly when running the test:

```python
(Pdb) request.method is not 'OPTIONS'
False
(Pdb) request.method != 'OPTIONS'
False
(Pdb) id(request.method)
38093424
(Pdb) id('OPTIONS')
38093424
```

Interestingly:

```python
(Pdb) id(request.method) == id('OPTIONS')
False
(Pdb) id(intern(request.method)) == id(intern('OPTIONS'))
True
```

ametaireau: So, python sometimes optimizes memory location for strings, but only sometimes :)

From [the python docs](https://docs.python.org/3/reference/datamodel.html#objects-values-and-types), I read that here we're comparing the id of the container with the id of the string, and so, boom.

> The value of some objects can change. Objects whose value can change are said to be mutable; objects whose value is unchangeable once they are created are called immutable. (The value of an immutable container object that contains a reference to a mutable object can change when the latter’s value is changed; however the container is still considered immutable, because the collection of objects it contains cannot be changed. So, immutability is not strictly the same as having an unchangeable value, it is more subtle.) An object’s mutability is determined by its type; for instance, numbers, strings and tuples are immutable, while dictionaries and lists are mutable.

simock85:
It's implementation-specific, but the interpreter is interning compile-time constants, but not run-time stuff, so `id(request.method)` -> compile-time const, is not equal to `id('OPTIONS')` -> run-time. If you manually `intern` run-time stuff, then you get the same object.

PR #281, opened 2015-03-11
Service.cors_supported_headers are now filtered by method

leplatrem:
Fix proposal of https://github.com/mozilla-services/cliquet/issues/54

*caution*, this is a violent change, breaking the Service API (from property to class method)

### DIFF
--- a/cornice/service.py
+++ b/cornice/service.py
@@ -413,17 +413,21 @@ class Service(object):
     def cors_enabled(self, value):
         self._cors_enabled = value
 
-    @property
-    def cors_supported_headers(self):
+    def cors_supported_headers(self, method=None):
         """Return an iterable of supported headers for this service.
 
         The supported headers are defined by the :param headers: argument
         that is passed to services or methods, at definition time.
         """
         headers = set()
-        for _, _, args in self.definitions:
+        for meth, _, args in self.definitions:
             if args.get('cors_enabled', True):
-                headers |= set(args.get('cors_headers', ()))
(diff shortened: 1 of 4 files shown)


leplatrem:
I suggest we also change ``cors_supported_methods`` and ``cors_supported_origins`` to be methods instead of properties.

cornice/service.py
```diff
@@ -413,17 +413,21 @@ def cors_enabled(self):
     def cors_enabled(self, value):
         self._cors_enabled = value
 
-    @property
-    def cors_supported_headers(self):
+    def cors_supported_headers(self, method=None):
```
ametaireau: When exposing functions with the first argument `method`, we've been following the `{methodname}_for` paradigm, maybe we should stick with this.

ametaireau: Also, that would let us be backwards compatible, exposing the properties to the people who where already using them. In this case, we should issue a deprecation warning and remove it in a later release.

ametaireau:
Otherwise, +1 with the idea.

cornice/service.py
leplatrem: I updated the PR, and generalized it to cors_support_credentials.

It looks weird when method param is omitted, but it's now consistent

leplatrem:
@ametaireau I now raise DeprecationWarning for previous properties/methods. Does it suit what you had in mind ?

leplatrem:
Thanks for your feedback, I changed the deprecation warnings to fit existing code in Cornice

ametaireau:
thanks!

PR #284, opened 2015-04-02
sphinx extension: display default values

MikaYuoadas:
When a default value exists for a schema attribute, display it instead of just
showing it is optional.

### DIFF
--- a/cornice/ext/sphinxext.py
+++ b/cornice/ext/sphinxext.py
@@ -6,6 +6,7 @@
 Sphinx extension that is able to convert a service into a documentation.
 """
 import sys
+import json
 from importlib import import_module
 
 from cornice.util import to_list, is_string, PY3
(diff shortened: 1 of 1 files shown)


cornice/ext/sphinxext.py
```diff
@@ -139,7 +140,12 @@ def _render_service(self, service):
                             if not attr.required or attr.description:
                                 temp += nodes.inline(text=' - ')
                                 if not attr.required:
-                                    temp += nodes.inline(text='(optional) ')
+                                    if attr.missing is not None:
+                                        default = json.dumps(attr.missing)
+                                        temp += nodes.inline(
+                                            text='(default: %s) ' % default)
+                                    else:
+                                        temp += nodes.inline(text='(optional) ')
```
ametaireau: wow, this makes this line too long, we should maybe refactor that a bit ;)

ametaireau:
Otherwise that looks good to me, thanks!

MikaYuoadas:
Ok, I edited the commit to make it pep8 compliant.

PR #285, opened 2015-04-02
Create swagger.py

hamx0r:
Here is a fully functional swagger spec generator which extracts documentation from
1) Pyramid Services (which may come from Cornice `@resources`) to generate Swager `paths`
2) Pyramid `@view`s if they contain Colander validation `schema`s to generate Swagger `parameters`
3) SQLAlchemy `model`s to generate Swagger `responses`

In addition, there are two decorators for one to explicitly add any Swagger item to a 'path' or 'operation':
```
@swagger.operation(parameters=[{
        "name": "car_id",
        "description": 'CSV list of cars',
        "required": True,
        "type": "string",
        "in": "path",
        "collectionFormat": "csv"
    }, {"name": "Range",
        "description": 'Range (ie "items=500-999") of flights requested',
        "required": False,
        "type": "string",
        "in": "header"}],"""
    @swagger.operation(responses=dict({"200": {"description": "CSV Body", "schema": {"type": "string"},
                            'headers': {'Content-Range':
                            {'description': 'Items in current page with total items availble (items 0-499/20000)',
                            'type': 'string'}}}}, **responses.unauthorized))#,
    @view(schema=validators.CarCSVGet)
    def get(self): ...
```
Here, the `@view(schema=` part will cause generation by the underlying Colander schema being inspected (ie our `validators.CarCSVGet` class).  Similarly, it may make sense for one to define `responses` and `parameters` in other files, and import them for use like so:
```
@swagger.operation(parameters=parameters.cars_get, responses=responses.cars_get)
```


ametaireau:
Nice, I would like to have tests for it though and documentation, but I love the feature :)

hamx0r:
Sure!   What's an example of a test you can recommend I follow?  What
format would you like documentation?  a .md file?

ametaireau:
For documentation, we're using the `rst` format, you can find out examples in the docs folder.

For tests, I believe what we want is to test we can generate the correct swagger definitions from a service definition, so it would be something among the lines of:

1. Define your service(s);
2. Pass them into your swagger convertor;
3. Check the result is what you would expect.

Also, It would make sense to test non-trival cases (colander, docstrings, etc).

hamx0r:
Here's some documentation i just finished.  I'll write a test in the next few days.  
https://github.com/hamx0r/cornice/blob/patch-1/cornice/ext/swagger.md

ametaireau:
Nice! You could include that in the docs, it's in RST format.

hamx0r:
Oops!  Done: https://github.com/hamx0r/cornice/blob/patch-1/docs/source/swagger.rst

ametaireau:
great. now the tests are we'll be there faster than we think :+1:

cornice/ext/swagger.py
```diff
@@ -0,0 +1,370 @@
...
+    json_model['required'] = list(set(required))
+    json_model['title'] = str(title)
+    return json_model
+
+#TODO make multidict2matchdict for schema objects.
+
+##############################
+#  SWAGGER SPEC GENERATORS   #
+##############################
+def generate_swagger_spec(services, title, version, **kwargs):
```
ametaireau: Oh wow, this is a big function with a lot of things inside. We should make it smaller grained :)

cornice/ext/swagger.md
```diff
@@ -0,0 +1,211 @@
+# Overview
```
ametaireau: This file shouldn't be here.

hamx0r:
I removed swagger.md.   Regarding breaking up generate_swagger_spec, it sounds fine to break it up.  FYI: i'm on travel for a few months and will be slow to make progress on this PR (need to write a test and break up the large function).

leplatrem:
Hi @hamx0r!

Would you still be motivated by this initiative?

Let us know, we may find a way to catch on and land this feature!

hamx0r:
Yes! It's the overhead that's making it hard to contribute. I'll update my
pull request with my latest code. Hopefully others can help with test
writing and breaking my code into smaller functions.

leplatrem:
Closing in favor of #340

PR #286, opened 2015-04-08
"Simply" is redundant and well, not accurate.

ijames:
I believe we should strike "simply", "easily", and "obviously" from all tutorials or presentations.  :-)  I'm working through cornice now and just thought I'd provide a tad of cleanup.


ametaireau:
Obviously, this is simply better.

ijames:
Wait, is it obvious?  I'm confused.

ijames:
Whoo hoo!

PR #287, opened 2015-04-10
Fix #264

### LINKED ISSUE #264: ACL and permission not work
I'm trying to set permissions for my api, but cornice ignore acl or permission attribute. Simple example(I want to Deny api for all, just for test):

```python
import json

from cornice.resource import resource, view
from pyramid.security import (ALL_PERMISSIONS, Allow, authenticated_userid,
                              Deny, DENY_ALL, Everyone)
...
from ..models import MyModel


def Test(request):
    return [
        (Deny, Everyone, ('get', 'collection_get')),
        DENY_ALL
    ]


@resource(collection_path='/foo', path='/foo/{uuid}')
class Foo(object):

    def __init__(self, request):
        self.request = request
        self.obj_query = DBSession.query(MyModel)

    def collection_get(self):
        objs = [item.__json__() for item in self.obj_query.all()]
        return {'mymodel': json.dumps(objs)}

    @view(renderer='json', acl=Test, permission='get')
    def get(self):
        ...
```

circlingthesun:
This fixes #264. The issue was that the route for a service is added once for each method. Each time the original route is overridden with a new root factory. The default options method has no ACL attached so it ends up overwriting the route with a root factory with no ACL. The way routes are added needs to be rethought. This is fixes the problem for the interim but the order in which routes are added could still result in issues.

### DIFF
--- a/cornice/service.py
+++ b/cornice/service.py
@@ -290,20 +290,20 @@ class Service(object):
 
         if 'acl' in kwargs:
             kwargs.pop('acl')
-            msg = "Warning: Using acl in method definitions is deprecated." +\
-                "Use service or resource instead."
+            msg = ("Warning: Using acl in method definitions is deprecated."
+                   "Use service or resource instead.")
             warnings.warn(msg, DeprecationWarning)
 
         if 'factory' in kwargs:
             kwargs.pop('factory')
-            msg = "Warning: Using acl in method definitions is deprecated." +\
-                "Use service or resource instead."
+            msg = ("Warning: Using acl in method definitions is deprecated."
+                   "Use service or resource instead.")
             warnings.warn(msg, DeprecationWarning)
 
         if 'traverse' in kwargs:
             kwargs.pop('traverse')
-            msg = "Warning: Using traverse in method definitions is deprecated." +\
-                "Use service or resource instead."
+            msg = ("Warning: Using traverse in method definitions is "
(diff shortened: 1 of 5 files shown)


circlingthesun:
Overwriting the route for every view is not ideal. Because the ACL/Factory/Traverse can only be specified when defining the route, it does not make sense to have these options on view decorators. I have therefore deprecated the options in this PR. I've also added a collection_acl option that will allow one to specify a separate ACL for collections when using resources. Test have also been updated.

cornice/service.py
```diff
@@ -287,6 +287,25 @@ def add_view(self, method, view, **kwargs):
         :param **kwargs: additional configuration for this view,
                         including `permission`.
         """
+
+        if 'acl' in kwargs:
+            kwargs.pop('acl')
+            msg = "Warning: Using acl in method definitions is deprecated." +\
```
ametaireau: nit: use parenthesis for joining strings rather than adding them together.

ametaireau:
Good catch. It emphasize my non-knowledge of routes vs view decorators at the time I wrote this.

The fix looks legit to me. I think this will need a major bump for the next release though.

cornice/service.py
circlingthesun: Fixed

circlingthesun:
Cool, looking forward to the next release :)

PR #288, opened 2015-04-13
Add option to disable schema request binding

leplatrem:


cornice/tests/test_schemas.py
```diff
@@ -204,6 +204,18 @@ def test_colander_bound_schema_rebinds_to_new_request(self):
             field = the_schema.get_attributes(request=other_dummy_request)[3]
             self.assertEqual(field.validator.choices, ['c', 'd'])
 
+        def test_colander_request_binding_can_be_disabled(self):
+            dummy_request = {'x-foo': 'version_a'}
+            the_schema = CorniceSchema.from_colander(ToBoundSchema)
+            field = the_schema.get_attributes(request=dummy_request)[3]
+            # Deferred are resolved
```
ametaireau: Should this be two tests?

ametaireau:
We should add this to the documentation also, otherwise this looks good to me.

cornice/tests/test_schemas.py
leplatrem: +1

leplatrem:
I wanted to mention it in the documentation, but found not appropriate existing section :/

ametaireau:
There is no specific docs about the colander request binding yet, but we could add it to http://cornice.readthedocs.org/en/latest/validation.html#using-colander

ametaireau:
Alternatively, adding a docstring could make this more explicit

leplatrem:
I chose to rewrite this part of the documentation :)

ametaireau:
Thanks!

PR #293, opened 2015-04-28
Allow arbitrary path prefixes in Cornice resource.

ktomala:
This pull request allows arbitrary path prefixes in Cornice **resource**. It can be very useful for grouping related methods, sub-resources, etc. Changes to **resource** decorator drop the notion of having only *path* and *collection_path* and allows for any number of prefixed paths. 

For example if you want to create sub-resource for *users* collection that will allow for example issuing methods or manipulating depended collections (like e.g. *friends*) you can create a resource like this:
@resource(collection_path='/users', path='/users/{id}', friends_path='/users/{id}/friends', friend_path='/users/{id}/friends/{name}')
This is much cleaner than building another **resource** that has to utilize only *collection_path* and *path* and sometimes may be useful for defining actions on resource instance like e.g. '/users/{id}/activate'. I know that some say this is sometimes not very RESTful way of doing things, and should be done with path e.g. '/userActivate/{id}', but in my opinion allowing defining sub-resources and additional paths within the same class logically groups the code for closely related HTTP methods. 

This change shouldn't  in any way break existing code, but simply allows extending Cornice resource beyond what was possible until now.

### DIFF
--- a/cornice/resource.py
+++ b/cornice/resource.py
@@ -18,41 +18,43 @@ def resource(depth=1, **kw):
     will be used as such. You can also prefix them by "collection_" and they
     will be treated as HTTP methods for the given collection path
     (collection_path), if any.
-    
+
     Here is an example::
 
         @resource(collection_path='/users', path='/users/{id}')
-        
+
     You can also add any other prefixed path like e.g. "activate_path".
     This may be useful when defining sub-resources.
 
     Here is an example::
 
-        @resource(collection_path='/users', path='/users/{id}', 
+        @resource(collection_path='/users', path='/users/{id}',
                   activate_path='/users/{id}/activate')
     """
     def wrapper(klass):
         services = {}
 
-        path_keys = [x[:len(x)-len('path')] for x in list(kw) if x.endswith('_path')]
+        path_keys = [x[:len(x)-len('path')] for x in list(kw)
+                     if x.endswith('_path')]
         if len(path_keys) > 0:
             prefixes = tuple(path_keys + [''])
         else:
             prefixes = ('',)
-        
+
         def filter_prefixes(prefix, x):
             """Filters out keys for prefixes other than specified prefix.
             """
             other = set(prefixes) - set(['', prefix])
             return len([y for y in other if x.startswith(y)]) == 0
-        
+
         for prefix in prefixes:
 
             # get clean view arguments
             service_args = {}
-            
-            filter_other = filter(functools.partial(filter_prefixes, prefix), list(kw))
-            
+
+            filter_other = filter(functools.partial(filter_prefixes, prefix),
+                                  list(kw))
+
             for k in filter_other:
                 if k.startswith(prefix):
                     service_args[k[len(prefix):]] = kw[k]
--- a/cornice/resource.py
+++ b/cornice/resource.py
@@ -8,6 +8,7 @@ try:
     VENUSIAN = True
 except ImportError:
     VENUSIAN = False
+import functools
 
 
 def resource(depth=1, **kw):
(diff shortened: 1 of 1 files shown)


ktomala:
Added another commit to fix flake8 issues.

ametaireau:
Hi, thanks for the pull request!

As a general rule, pull requests to cornice should always come with tests and documentation and cannot be merged otherwise.

Without these tests, I fail to see what's the use case you're trying to serve here. I feel introducing other paths to a resource definition is kinda out of scope: a resource only has a path and a collection_path, and the whole class is about to ease the creation of RESTful resources. In case you add other paths, this spirit seems broken to me.

I'm not saying you shouldn't do it, but there are probably other ways than including that into cornice that should let you do that, and using the basic service class should let you do that in a straight forward way.

ktomala:
Hi Alexis.
I will do tests and documentation for this pull request. This pull request in my opinion allows you to group relevant methods for one resource that otherwise shouldn't exist out of scope for the defined resource. Think about sub-resources that are possible within RESTful scope. I.e. if you have depended resources that cannot otherwise exist without the main resource, you can define additional paths that serves this exact purpose. 

Let's suppose you have "Users" resource. The "User" has it's depended resources (sub-resources) like "Friends". If you make general methods for querying this user from the data source (db or otherwise) and the subresource "Friends" depends on selecting proper "User" by means of "id" field passed in matchdict, this could ease the development of methods for such resource. If you have a path="/users/{id}" and friends_path="/users/{id}/friends" you can then create method for selecting User object by it's "id" field and then query "friends" field or relation on the User.id == request.matchdict['id'] instance. 

I am aware that you can do this with Service classes, but it seems to me a little bit counterintuitive and leaves you with reimplementing select method for User object in each Service instance class that is relevant to all it's sub-resources.

dairiki:
In the user-friends use case, I would have — perhaps, naively — written things like:
```python
@resource(collection_path='/users', path='/users/{id}', ...)
class UserResource(object):
    ...

@resource(collection_path='/users/{id}/friends', path='/users/{id}/friends/{name}', ...)
class UserFriendResource(object):
    ...
```
I.e. a *user* is a resource, and a *user's friend* is a different resource, related though it may be.

What is to be gained by balling all these resources together in a single class?

ktomala:
Perhaps in the user-friends use case it may not be as obvious as to what the advantages are, but assume for a second that you have an interface tailored to the data store. This interface based on Cornice Resource has built-in methods for automatic querying specific objects from the data store that are dependent upon URI path. So when you query "/user/{id}" you get User object from the data store based upon it's id. If the interface is built on the principle that it automatically queries specified objects and you have sub-resource that is related to the parent object, i.e. Friend, the interface knows that it has to first query User (i.e. make checks for it's id, perform data store fetch, transform data, etc.) and then use this User object as an input to search for specific relation named "friends" and query the data store for this object. However you can imagine that you want to first check for User object existence in the data store (essentially duplicating the code in the /user/{id} GET method), before you query for specific Friend. This interface will then perform the parent object check, query and transformation transparently for you, without code duplication. This is where I think it is useful to have it in one class. 

Bear in mind that what I have stated as an example, should be treated as one - an example. Maybe not very useful or self-explanatory. Of course. You can create another class inherited from Resource that implements this functionality, or you can create many different resources like you have shown before, but then you loose coherence and it may not be as straightforward or even possible to use an interface like I have mentioned above. Of course that is not a problem if this commit gets rejected from Cornice, because I can simply create another resource-like class for my needs, that will implement this functionality and work properly with this interface.

However, there's another use case, which may seem weak, but not every RESTful interface is pretty or even designed properly according to principles. Let's suppose you have a legacy interface, that is ported to Cornice and the way the interface was designed is that it has some RPC-like methods on a specific resource. This is only an example, but sometimes you have to define additional methods that are bound to the resource state (during a transaction or another kind of process) like for example /users/{id}/activate or perhaps /vm/{id}/shutdown, etc. 

You could use a Service for it, but IMO that looks messy and may be easily omitted by people reading the code. Also there's a problem with resource state dependent actions, that you'd have to create a state container shared between the resource and service, which again looks messy, introduces sometimes unnecessary abstraction layer and may result in performance impact. Of course you can decorate your resource class methods with Service decorators instantiated in __init__ method of the resource, but this looks messy and inconsistent.

However if you opt to reject this commit, that should not be a problem for me, because I will simply reimplement this functionality in an inherited class.

ametaireau:
I fear designing our interfaces with use cases such as legacy code / interfaces isn't a path we want to follow.

Also, given the fact that the resource methafor has a bunch of known issues and that we have plans to make it evolve in the future, I'm not sure extending it to allow more (corner) cases is a good idea for now.

If Cornice would be in your way here, I would of course accept a pull request to change this, but as it's possible to have this code elsewhere of Cornice, I believe it's wiser to do so.

That being said, thanks for taking the time to contribute and to start the discussion. I know it's hard to see requests rejected but it's nothing to take personally, I just feel it's not the right thing to do.

Thanks!

ktomala:
No offence taken. However, could we include this as an extension or should I create another repository just for this change, renaming the class AbstractResource or something like this?

ametaireau:
Another repository is probably the best way to handle this. Post it here for later reference!

PR #294, opened 2015-04-28
Add validate_matchdict and validate_wtforms_schema to Cornice schemas.

ktomala:
This pull request adds two useful validation methods:

* validate_matchdict - which allows simple checking types of arguments passed to view by request.matchdict
* validate_wtforms_schema - which allows validating JSON data through WTForms schema. Nested JSON dictionary validation requires importing and initializing wtforms_json extension within the module where WTForms schemas are defined.

### DIFF
--- a/cornice/ext/wtforms.py
+++ b/cornice/ext/wtforms.py
@@ -1,0 +1,52 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+def validate_wtforms_schema(form_schema, with_matchdict=False):
+    """Validates JSON data through WTForms schema.
+
+    For this function to properly work you have to install wtforms_json
+    package and add this code::
+
+        import wtforms_json
+        wtforms_json.init()
+
+    to the module where you define WTForms schemas, just after WTForms
+    imports.
+
+    Here's an example usage::
+
+        @view(
+            ...
+            validators=( validate_wtforms_schema(MyFormSchema), ),
+            ...
+        )
+
+    This function also has a keyword argument *with_matchdict* set
+    as default to False. Setting *with_matchdict* to True allows
+    to update JSON dictionary with values passed through
+    request.matchdict. It may be useful when you pass e.g. *id*
+    within your request.matchdict and you want to merge keys from
+    it to the JSON dictionary for further processing e.g.
+    to update SQLAlchemy model.
+    """
+    def wrapper(request):
+        try:
+            data = request.json
+        except Exception as e:
+            request.errors.add('body', 'exception', str(e))
+            return
+
+        if with_matchdict:
+            data.update(request.matchdict)
+
+        try:
+            form_obj = form_schema.from_json(data)
+        except Exception as e:
+            request.errors.add('form', 'exception', str(e))
+            return
+        if not form_obj.validate():
+            request.errors.add('form', 'validation', form_obj.errors)
+            return
+        request.form = form_obj
+    return wrapper
--- a/cornice/ext/wtforms.py
+++ b/cornice/ext/wtforms.py
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 
+
 def validate_wtforms_schema(form_schema, with_matchdict=False):
     """Validates JSON data through WTForms schema.
 
--- a/cornice/schemas.py
+++ b/cornice/schemas.py
@@ -181,7 +181,7 @@ def validate_matchdict(matchdict_types):
     """Validates types of request.matchdict through matchdict_types
     dictionary.
 
-    matchdict_types dictionary should have it's values defined as
+    matchdict_types dictionary should have its values defined as
     callable type. Here's an example::
 
         @view(
--- a/cornice/schemas.py
+++ b/cornice/schemas.py
@@ -198,5 +198,5 @@ def validate_matchdict(matchdict_types):
                 try:
                     typ(item)
                 except Exception as e:
-                    request.errors.add('matchdict', name, str(e))
+                    request.errors.add('path', name, str(e))
     return wrapper
--- a/cornice/schemas.py
+++ b/cornice/schemas.py
@@ -191,62 +191,12 @@ def validate_matchdict(matchdict_types):
         )
     """
     def wrapper(request):
-        md = request.matchdict
-        for k, v in matchdict_types.iteritems():
-            if k in md.keys():
-                item = md.get(k)
+        matchdict = request.matchdict
+        for name, typ in matchdict_types.iteritems():
+            if name in matchdict.keys():
+                item = matchdict.get(name)
                 try:
-                    v(item)
+                    typ(item)
                 except Exception as e:
-                    request.errors.add('param', 'exception', str(e))
-    return wrapper
-
-
-def validate_wtforms_schema(form_schema, with_matchdict=False):
-    """Validates JSON data through WTForms schema.
-
-    For this function to properly work you have to install wtforms_json
-    package and add this code::
-
-        import wtforms_json
-        wtforms_json.init()
-
-    to the module where you define WTForms schemas, just after WTForms
-    imports.
-
-    Here's an example usage::
-
-        @view(
-            ...
-            validators=( validate_wtforms_schema(MyFormSchema), ),
-            ...
-        )
-
-    This function also has a keyword argument *with_matchdict* set
-    as default to False. Setting *with_matchdict* to True allows
-    to update JSON dictionary with values passed through
-    request.matchdict. It may be useful when you pass e.g. *id*
-    within your request.matchdict and you want to merge keys from
-    it to the JSON dictionary for further processing e.g.
-    to update SQLAlchemy model.
-    """
-    def wrapper(request):
-        try:
-            data = request.json
-        except Exception as e:
-            request.errors.add('data', 'json', str(e))
-            return
-
-        if with_matchdict:
-            data.update(request.matchdict)
-
-        try:
-            form_obj = form_schema.from_json(data)
-        except Exception as e:
-            request.errors.add('form', 'exception', str(e))
-            return
-        if not form_obj.validate():
-            request.errors.add('form', 'validation', form_obj.errors)
-            return
-        request.form = form_obj
+                    request.errors.add('matchdict', name, str(e))
     return wrapper
--- a/cornice/schemas.py
+++ b/cornice/schemas.py
@@ -175,3 +175,78 @@ def validate_colander_schema(schema, request):
         for param in set(params) - set([attr.name for attr in attrs]):
             request.errors.add('body' if param in body else 'querystring',
                                param, msg % param)
+
+
+def validate_matchdict(matchdict_types):
+    """Validates types of request.matchdict through matchdict_types
+    dictionary.
+
+    matchdict_types dictionary should have it's values defined as
+    callable type. Here's an example::
+
+        @view(
+            ...
+            validators=( validate_matchdict( {'id': int} ), ),
+            ...
+        )
+    """
+    def wrapper(request):
+        md = request.matchdict
+        for k, v in matchdict_types.iteritems():
+            if k in md.keys():
+                item = md.get(k)
+                try:
+                    v(item)
+                except Exception as e:
+                    request.errors.add('param', 'exception', str(e))
+    return wrapper
+
+
+def validate_wtforms_schema(form_schema, with_matchdict=False):
+    """Validates JSON data through WTForms schema.
+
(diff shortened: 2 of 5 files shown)


cornice/schemas.py
```diff
@@ -175,3 +175,78 @@ def _validate_fields(location, data):
...
+            if k in md.keys():
+                item = md.get(k)
+                try:
+                    v(item)
+                except Exception as e:
+                    request.errors.add('param', 'exception', str(e))
+    return wrapper
+
+
+def validate_wtforms_schema(form_schema, with_matchdict=False):
```
ametaireau: This could be an extension instead of being part of cornice/schemas.py.

Extensions leave in `cornice/ext`.

cornice/schemas.py
```diff
@@ -175,3 +175,78 @@ def _validate_fields(location, data):
...
+    callable type. Here's an example::
+
+        @view(
+            ...
+            validators=( validate_matchdict( {'id': int} ), ),
+            ...
+        )
+    """
+    def wrapper(request):
+        md = request.matchdict
```
ametaireau: I would name this `matchdict` or `submited_data`.

cornice/schemas.py
```diff
@@ -175,3 +175,78 @@ def _validate_fields(location, data):
...
+        for k, v in matchdict_types.iteritems():
```
ametaireau: `for name, typ in`

cornice/schemas.py
ktomala: You mean the whole pull request or one of those two methods?

cornice/schemas.py
```diff
@@ -175,3 +175,78 @@ def _validate_fields(location, data):
         for param in set(params) - set([attr.name for attr in attrs]):
             request.errors.add('body' if param in body else 'querystring',
                                param, msg % param)
+
+
+def validate_matchdict(matchdict_types):
```
ametaireau: That's an interesting approach but I don't know if we want to have that in core cornice though. @leplatrem @tarekziade thoughts?

ametaireau: In any case, to have that included, it misses tests and docs :)

cornice/schemas.py
ametaireau: Just the last method.

cornice/schemas.py
ktomala: You mean the module for extension or method? I don't think this should be submitted_data, because this method relies upon matchdict which does not have anything to do with submitted_data (i.e. POST/PUT form data). Or are you referring to validate_wtforms_schema?

cornice/schemas.py
```diff
@@ -175,3 +175,78 @@ def _validate_fields(location, data):
...
+    """
+    def wrapper(request):
+        md = request.matchdict
+        for k, v in matchdict_types.iteritems():
+            if k in md.keys():
+                item = md.get(k)
+                try:
+                    v(item)
+                except Exception as e:
+                    request.errors.add('param', 'exception', str(e))
```
ametaireau: This is not how these errors are supposed to be used. This should be something like `request.errors.add(location, name, description)`

cornice/schemas.py
ktomala: Ok. I can add tests and docs for validate_matchdict method.

cornice/schemas.py
ametaireau: naming this `md` isn't explicit so renaming it should help.

ktomala:
Refactoring and cleanup for new validation methods. Added documentation for validate_matchdict and validate_wtforms_schema.

cornice/schemas.py
```diff
@@ -175,3 +175,28 @@ def _validate_fields(location, data):
         for param in set(params) - set([attr.name for attr in attrs]):
             request.errors.add('body' if param in body else 'querystring',
                                param, msg % param)
+
+
+def validate_matchdict(matchdict_types):
+    """Validates types of request.matchdict through matchdict_types
+    dictionary.
+
+    matchdict_types dictionary should have it's values defined as
```
leplatrem: typo: its values

ktomala: Thanks, fixed.

docs/source/validation.rst
```diff
@@ -232,6 +232,81 @@ like this::
...
+
+    from cornice import Service
+    import re
+	
+	FOO = {}
+	
+    foo = Service(name='foo', path='/foo/{id}')
+    
+    
+    def coerce_email(value):
```
leplatrem: Email may not be the most common use-case. What about UUID for example ?

docs/source/wtforms.rst
```diff
@@ -0,0 +1,137 @@
...
+that should be validated through WTForms schema. Our User will also have address
+defined, so we can show how to validate nested JSON schemas through WTForms.
+
+Let's start with importing required packages::
+
+    from cornice import Service
+    from cornice.ext.wtforms import validate_wtforms_schema
+    
+    from wtforms import Form, StringField, FormField, IntegerField
+	from wtforms.validators import AnyOf, Length, DataRequired, Optional, \
```
leplatrem: indentation

docs/source/validation.rst
ktomala: It's an example, but if you want I can add another for UUID.

docs/source/wtforms.rst
```diff
@@ -0,0 +1,137 @@
...
+    class UserUpdateSchema(UserSchema):
+        id = IntegerField(validators=[InputRequired()]
+    
+
+    userupdate = Service(name='userupdate', path='/user/{id}')
+
+    
+    @userupdate.put(
+   		content-type='application/json',
+        validators=validate_wtforms_schema(UserUpdateSchema, with_matchdict=True))
```
leplatrem: indentation could be improved here IMO

cornice/schemas.py
```diff
@@ -175,3 +175,28 @@ def _validate_fields(location, data):
...
+    """
+    def wrapper(request):
+        matchdict = request.matchdict
+        for name, typ in matchdict_types.iteritems():
+            if name in matchdict.keys():
+                item = matchdict.get(name)
+                try:
+                    typ(item)
+                except Exception as e:
+                    request.errors.add('matchdict', name, str(e))
```
leplatrem: I think the location is ``'path'``

cornice/schemas.py
leplatrem: I believe that it could be part of Cornice. I find it much more elegant to use a validator like this than [what we did in cliquet](https://github.com/mozilla-services/cliquet/blob/master/cliquet/resource.py#L233) for example.


Even if I don't like it too much, Cornice provides a way to validate several parts of the request using the ``location`` argument in a schema. Shouldn't we take advantage of it instead ?

```python
class Schema():
    id = colander.SchemaNode(colander.Int, location='path')
    phone = colander.SchemaNode(colander.String, location='body')
```

ktomala: The solution you have posted may be alright if you want to use Colander, however if you want to use other libraries for schema validation, this may not be the case. I think for some use cases my solution is much simpler and more performant.

ktomala:
Fixed indentiation in documentation.

docs/source/validation.rst
ametaireau: It might be enough to update this example with uuid rather than email.

ametaireau:
Nice! It just misses tests and we're good to go!

cornice/ext/wtforms.py
```diff
@@ -0,0 +1,53 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this file,
+# You can obtain one at http://mozilla.org/MPL/2.0/.
+
+
+def validate_wtforms_schema(form_schema, with_matchdict=False):
```
ametaireau: Since the module is already named "wtforms" I believe it's enough to name this `validate_schema`.

cornice/ext/wtforms.py
```diff
@@ -0,0 +1,53 @@
...
+
+    Here's an example usage::
+
+        @view(
+            ...
+            validators=( validate_wtforms_schema(MyFormSchema), ),
+            ...
+        )
+
+    This function also has a keyword argument *with_matchdict* set
```
ametaireau: Rather than defining the existance and use of this parameter in the docstring like that, we try to use the `:param with_matchdict` syntax. See https://github.com/mozilla-services/cornice/blob/master/cornice/service.py#L51-L52 for an example.

cornice/ext/wtforms.py
```diff
@@ -0,0 +1,53 @@
...
+    request.matchdict. It may be useful when you pass e.g. *id*
+    within your request.matchdict and you want to merge keys from
+    it to the JSON dictionary for further processing e.g.
+    to update SQLAlchemy model.
+    """
+    def wrapper(request):
+        try:
+            data = request.json
+        except Exception as e:
+            request.errors.add('body', 'exception', str(e))
```
ametaireau: I think I miss what we're trying to catch here, but exercised with a test I would probably understand better.

cornice/ext/wtforms.py
```diff
@@ -0,0 +1,53 @@
...
+            return
+
+        if with_matchdict:
+            data.update(request.matchdict)
+
+        try:
+            form_obj = form_schema.from_json(data)
+        except Exception as e:
```
ametaireau: Do we want to catch the global `Exception` here or a more defined one?

docs/source/validation.rst
```diff
@@ -232,6 +232,111 @@ like this::
...
+Sometimes it can be useful to validate types of variables passed through 
+**request.matchdict** e.g. when you have a Service method that depends upon 
+proper type of this variable.
+
+For example in CRUD-like services you often define methods (GET/PUT/DELETE) 
+whose URI includes an **id** that is used for querying the object. To ensure
+that your code doesn't break, you can validate coercion of the specified 
+**request.matchdict** variable to given type. For example::
+
+    from cornice import Service
```
ametaireau: Please import the validate_matchdict from `cornice.schemas` here.

ametaireau: Well, actually I think the right location would be `cornice/validators.py` anyway.

docs/source/validation.rst
ametaireau: Ah okay I've seen the UUID exemple below. That's okay for me.

docs/source/wtforms.rst
```diff
@@ -0,0 +1,137 @@
...
+    pip install wtforms-json
+    
+Suppose you want to create a Service that allows you to send JSON data for User
+that should be validated through WTForms schema. Our User will also have address
+defined, so we can show how to validate nested JSON schemas through WTForms.
+
+Let's start with importing required packages::
+
+    from cornice import Service
+    from cornice.ext.wtforms import validate_wtforms_schema
```
ametaireau: validate_schema

ametaireau:
Neat! I'm looking forward to see this included.

ktomala:
I will apply fixes to what you have stated above later this week since I am a little bit busy right now. As for the tests, I am not sure what kind of tests would you like to see? Any thoughts on this?

cornice/ext/wtforms.py
ktomala: This try/except is when data passed through request is malformed (i.e. JSON errors like e.g. single quote instead of double, ommited/excess comma, etc.)

cornice/ext/wtforms.py
ktomala: I believe we should catch global Exception, because we are depending on external libraries here (WTForms, wtforms-json), which may change in the future, so better safe than sorry.

docs/source/validation.rst
ktomala: Ok. I'll move this code. Currently cornice/validators.py is almost empty, so I thought a better suited location would be schema.py.

ametaireau:
For the tests, we're looking for unit-tests that prove this is working as intended. @ktomala thanks for taking some time for this, appreciate it.

ktomala:
Thanks for the information @ametaireau. I will try to do the unit-tests as soon as possible. Are there any preferred testing frameworks (like e.g. nose) or should I use built-in Python tests?

ametaireau:
We already have a set of tests (look for the tests folder). You can run the tests using `tox` or `python setup.py test`.

tisdall:
@ametaireau - `python setup.py test` doesn't work...  see #308

ametaireau:
Ah yes true. My previous statement still applies. Need some time :)

> It seems that something we really miss for cornice is a "how to contribute" documentation page, explaining how to run the tests, etc.

cornice/schemas.py
leplatrem: I agree with you!

cornice/ext/wtforms.py
leplatrem: Ok, maybe. But I would at least add a comment (and a test) for this :)

leplatrem:
With the new API on master, it becomes very easy to implement validation with any kind of library. Have a look! I'm closing this for now, but it could be revamped with the new approach! Thanks again!

PR #295, opened 2015-05-04
Import reload for Python 3

JohnBrodie:
Python 3 does not have the `reload` built-in.  As of Python 3.4, `reload` is available in `importlib`, see
https://docs.python.org/3/library/importlib.html#importlib.reload.

A test is included that shows the breakage/fix.  I'll be happy to make any additional changes requested.


ametaireau:
Thanks!

PR #303, opened 2015-06-04
Fixes #302 : add Access-Control-Allow-Headers

### LINKED ISSUE #302: Cors Access-Control-Allow-Headers
Recently Google Chrome started complaining:

```
XMLHttpRequest cannot load http://localhost:8000/v1/people. Request header field Content-Type is not allowed by Access-Control-Allow-Headers.
```

bkuberek:


bkuberek:
There is an issue with this PR in that it is not accounting for the `Access-Control-Request-Method`. I'm looking into that now.

bkuberek:
Ok i realized this PR doesn't solve the problem. The problem is that when we define `option` method in resource, it overrides the `get_cors_preflight_view` and therefore those headers are not set. We need to figure out a way to define `collection_options` and `options` in our resources without losing the headers generated in `get_cors_preflight_view`

ametaireau:
Some thoughts about that:

- A simple way to do so would be to decorate your view with the preflight view directly;
- Otherwise, by default, if it's an OPTION that's being defined, it might make sense to decorate it with the preflight view before registering the two. This behaviour should be configurable (with a `add_cors_preflight` flag that could be set to either `True` or `False`).

PR #308, opened 2015-06-10
make it so `python setup.py test` works

tisdall:


tisdall:
k..  it seems from the commit history that you had this already but @Natim took it out.  I don't really understand why, so I'll close this.

ametaireau:
To run the tests, we prefer to use tox in order to have all the supported versions tested.

ametaireau:
It seems that something we really miss for cornice is a "how to contribute" documentation page, explaining how to run the tests, etc.

Natim:
The problem I see with ``python setup.py test`` is that it install test_dependencies in the local directory rather than using the virtualenv, also as @ametaireau mentionned, using tox test matrix helps us test pep8 as well as multiple Python version at once.

tisdall:
@Natim I always run the `python` that's part of my virtualenv.

My point is that you can have both working.  In other projects I usually see the tox ini calling `python setup.py test` for each of the different versions.  That's what the code base seemed to be doing before, but now it isn't.

It's mostly just preference, though...  The other method restricts all the dependency definitions to the setup.py and then there's none in the tox.ini.  Also, when I ran `python setup.py test` it just installed stuff and then did nothing.  It's just what _I'm_ used to, though.

Issue #307, opened 2015-06-10
schema validation requires MappingSchema


tisdall:
This is related to this posting on SO:  http://stackoverflow.com/questions/30736627/cornice-schema-validation-with-colanderalchemy/30761861#30761861

Essentially validation will only work if the colander schema passed in is a direct subclass of `MappingSchema`.  Using something like `SchemaNode(typ=Mapping)` should work just fine, but instead it throws `cornice.schemas.SchemaError: schema is not a MappingSchema`.

ametaireau:
Yeah, we definitely should do some duck typing here rather than trying to do a type check. I'm not sure from the thread on Stack Overflow: do you want to work on a fix?

tisdall:
If you have time, I'd prefer it if you did.  I'm a little swamped.  I'm also just starting to figure out how Cornice works. ^_^

I'm going to fix ColanderAlchemy so it implements `schema_type`, but Cornice should probably be using `typ` as `schema_type` isn't really supposed to be used like that.  I filed an issue at https://github.com/Pylons/colander/issues/229 as that really is a hole waiting for other people to fall in.

tisdall:
Okay, on second thought, I'm not going to make the change to ColanderAlchemy.  The `schema_type()` is a static method so it has no access to the actual instance.  When you use ColanderAlchemy it assigns a value to `unknown` of `Mapping` and schema_type won't ever see that.  I think I could make it so `schema_type()` is no longer a static method, but then it's getting rather hackish.

I'll put together a PR to properly use the `typ` instead of `schema_type()` in Cornice.

tisdall:
I'm going to close this issue as PR #309 fixes this issue.

PR #309, opened 2015-06-12
Fix schema typing

tisdall:
fixes: #307

Unfortunately I didn't have the time to fully understand the code and try to do duck typing.  Right now it uses the `unknown` property that only really exists on the `Mapping` type.  The major thing is a SchemaNode(Mapping()) will now work.

### DIFF
--- a/cornice/schemas.py
+++ b/cornice/schemas.py
@@ -95,14 +95,14 @@ class CorniceSchema(object):
 
 def validate_colander_schema(schema, request):
     """Validates that the request is conform to the given schema"""
-    from colander import Invalid, Sequence, drop, null, MappingSchema
+    from colander import Invalid, Sequence, drop, null, Mapping
 
-    schema_type = schema.colander_schema.schema_type()
+    schema_type = schema.colander_schema.typ
     unknown = getattr(schema_type, 'unknown', None)
 
-    if not isinstance(schema.colander_schema, MappingSchema):
-        raise SchemaError('schema is not a MappingSchema: %s' %
-                          type(schema.colander_schema))
+    if not isinstance(schema_type, Mapping):
+        raise SchemaError('colander schema type is not a Mapping: %s' %
+                          type(schema_type))
 
     def _validate_fields(location, data):
         if location == 'body':
@@ -167,7 +167,7 @@ def validate_colander_schema(schema, request):
     _validate_fields('querystring', qs)
 
     # validate unknown
-    if schema.colander_schema.typ.unknown == 'raise':
+    if unknown == 'raise':
         attrs = schema.get_attributes(location=('body', 'querystring'),
                                       request=request)
         params = list(qs.keys()) + list(body.keys())
--- a/cornice/schemas.py
+++ b/cornice/schemas.py
@@ -97,6 +97,9 @@ def validate_colander_schema(schema, request):
     """Validates that the request is conform to the given schema"""
     from colander import Invalid, Sequence, drop, null, Mapping
 
+    # CorniceSchema.colander_schema guarantees that we have a colander
+    #  instance and not a class so we should use `typ` and not
+    #  `schema_type()` to determine the type.
     schema_type = schema.colander_schema.typ
     unknown = getattr(schema_type, 'unknown', None)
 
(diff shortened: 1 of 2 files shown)


ametaireau:
> The major thing is a SchemaNode(Mapping()) will now work.

This really is a problem, since it would break Cornice in an incompatible way with previous versions. We might need to find a better mechanism to decide if we should apply the validation.

For instance, it might make sense to be able to override this with a property of the passed schema, or pass a callable that should be called.

tisdall:
I don't understand.  Here's the definition of a MappingSchema in colander:
```
class Schema(SchemaNode):
    schema_type = Mapping

MappingSchema = Schema
```

It's just a `SchemaNode` with `schema_type() is Mapping()`.  `schema_type()` is called on `__init__` if no `typ` is provided.

`MappingSchema` is just a shortcut class if you don't want to do `SchemaNode(Mapping())`.  They're equivalent.

Without this change, it's impossible to use a `SchemaNode` which is the basic building block of all colander schema.  Also, I don't see how this change would break any functionality as it just removes erroneously throwing an error on `SchemaNode` instances that would otherwise work fine.

(maybe the confusion lies with my changes to the tests...  the tests pass with or without my changes.  The changes are just to make things clearer and more consistent.)

Additionally, colander has the "feature" of allowing you to do the following:
```
>>> import colander
>>> x = colander.MappingSchema(colander.Sequence())
>>> x.typ
<colander.Sequence object at 0x7fb908fc6208>
>>> x.typ.unknown
  File "<stdin>", line 1, in <module>
AttributeError: 'Sequence' object has no attribute 'unknown'
>>> 
```

So, testing if a class is an instance of a `MappingSchema` doesn't mean that it actually has a `Mapping` type. (https://github.com/Pylons/colander/pull/168)  So, you're really testing the wrong thing with checking if a schema is an instance of a `MappingSchema`.

ametaireau:
Thanks for working on this fix! I actually misread "now" with "not", sorry about that.

This fix will break cornice because the test isn't done in the same way. I know at least about one application that relies on the `schema_type` method.

So, here are a few things we should add to the this PR before merging:
- Tests! We need to showcase what this fixes, in order to avoid future problems when modifying the code;
- In case we will fail, we should probably check both `typ` and the `schema_type` before failing, since both are used in different projects.

tisdall:
The issue is, though, is `schema_type()` (despite the name) is not intended to tell you the type of the schema.  (See https://github.com/Pylons/colander/issues/229#issuecomment-111119270)  The only purpose/use of `schema_type()` is for subclassing `SchemaNode` and defining a default type to use if you don't supply one directly.  However, there's nothing restricting a subclass from just using a different type and ignoring `schema_type()`

example:
```
>>> import colander
>>> x = colander.MappingSchema(colander.Sequence())
>>> x.typ
<colander.Sequence object at 0x7ff71c745390>
>>> x.schema_type
<class 'colander.Mapping'>
```

In the above example, `x` should be treated as a schema with type `Sequence` despite what `schema_type` says (or what class it's subclassing from).

Also, `schema_type` is a static method that returns a new instance of a type, so with `Mapping` it won't return the proper value of `unknown`.

```
>>> x = colander.MappingSchema(colander.Mapping(unknown="raise"))
>>> x.typ
<colander.Mapping object at 0x7ff71c745358>
>>> x.typ.unknown
'raise'
>>> x.schema_type
<class 'colander.Mapping'>
>>> x.schema_type().unknown
'ignore'
```

So...  anything using `schema_type` for anything other than providing a default `typ` with sub-classing is just doing it wrong.  I think the only use for checking `schema_type` is if Cornice accepts schema classes and not instances (I'll have to defer to you if that happens... I'm not really sure).


I'll see what I can do about tests...

tisdall:
Okay, the call to `CorniceSchema.colander_schema()` guarantees that it's an instance and not a class so we should ignore `schema_type` there.  I'll add a comment to that effect to make sure it's clear.

tisdall:
Okay, I think this covers it...

tisdall:
Would you like me to make a note in CHANGES?  Is there anything else missing on this PR?

ametaireau:
/me summons our very own colander expert @leplatrem for a review :)

leplatrem:
I don't see anything wrong, and agree with @tisdall comments! it's indeed cleaner!

ametaireau:
Merged! Thanks :)

PR #312, opened 2015-06-15
make test less dependent on pyramid version

tisdall:
Running my own tests I got a failure on `test_basic_resource` using py34 and pyramid==1.5.7.  The message I got was this:
```
======================================================================
FAIL: test_basic_resource (cornice.tests.test_resource.TestResource)
----------------------------------------------------------------------
  File "/home/tim/cornice/cornice/tests/test_resource.py", line 119, in test_basic_resource
    self.assertEqual(resp.body, b'test({"name": "gawel"});', resp.body)
AssertionError: b'/**/test({"name": "gawel"});' != b'test({"name": "gawel"});' : b'/**/test({"name": "gawel"});'

----------------------------------------------------------------------
```

Instead of adding another branch on the if statement for that particular version, I changed the test to capture what we're really looking for...  It tests that the string contains 'test({"name": "gawel"})'.

I'm a little confused as to why tox fetches the alpha version of pyramid 1.6 instead of using the latest version found at https://pypi.python.org/pypi/pyramid (which is 1.5.7).  I guess because it's the most recent version found at https://pypi.python.org/simple/pyramid/ .

### DIFF
--- a/cornice/tests/test_resource.py
+++ b/cornice/tests/test_resource.py
@@ -104,19 +104,13 @@ class TestResource(TestCase):
         testing.tearDown()
 
     def test_basic_resource(self):
-        from pkg_resources import parse_version, get_distribution
-        current_version = parse_version(get_distribution('pyramid').version)
-
         self.assertEqual(self.app.get("/users").json, {'users': [1, 2]})
 
         self.assertEqual(self.app.get("/users/1").json, {'name': 'gawel'})
 
         resp = self.app.get("/users/1?callback=test")
 
-        if current_version < parse_version('1.5a4'):
-            self.assertEqual(resp.body, b'test({"name": "gawel"})', resp.body)
-        else:
-            self.assertEqual(resp.body, b'test({"name": "gawel"});', resp.body)
+        self.assertIn(b'test({"name": "gawel"})', resp.body, msg=resp.body)
 
     def test_accept_headers(self):
         # the accept headers should work even in case they're specified in a


dairiki:
FWIW, the `--pre` argument fed to `pip` in the `install_command` in `tox.ini` is, I think, what is causing the pre-release version of pyramid to be installed.

tisdall:
It'd be nice to test against the lastest versions of each major version of pyramid.  So, both 1.5.7 and 1.6a .  I guess that could be done with further tox environments, but then should that be done for every version of python too?  :)

ametaireau:
This looks good to me.

tisdall:
@ametaireau - should we remove the `--pre` argument in tox so the tests are run against the latest stable version instead of pre-releases?

ametaireau:
I find it useful to be able to test on pre-releases, isn't it?

tisdall:
@ametaireau - as I said above, I think it'd be nice to test against the latest version of each major release of pyramid.  However, if I had to choose which version is the most important to test against, I'd say it'd be the most recent stable release, no?

You could modify the tox file to test against both the pre-release and the latest stable (maybe those two versions of pyramid are sufficient).

ametaireau:
Yes that would make sense.

elemoine:
> You could modify the tox file to test against both the pre-release and the latest stable (maybe those two versions of pyramid are sufficient).

Makes sense to me as well. I like the idea of testing against pre releases to catch issues as early as possible.

PR #318, opened 2015-07-07
FIX tests that depends from jsonp response format

surabujin:
In pyramid version 1.5.7 JSONP renderer format was changed - "/**/" prefix was
added. Because of this our tests that check jsonp response become broken.
Extend JSONP related tests to consider that update.

### DIFF
--- a/cornice/tests/test_resource.py
+++ b/cornice/tests/test_resource.py
@@ -113,10 +113,13 @@ class TestResource(TestCase):
 
         resp = self.app.get("/users/1?callback=test")
 
-        if current_version < parse_version('1.5a4'):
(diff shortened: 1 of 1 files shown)


tisdall:
I fixed this in #312 and it's waiting someone to look at it.  I made it to test for the functional portion of the JSONP response instead of checking for every "/**/" or ";" depending on the version installed.

surabujin:
As for me, exact match/check is more correct variant. So better check some version-dependent variants, than use "fussy" match/check.

ametaireau:
This is now fixed in #312. Thanks to both of you :)

Issue #319, opened 2015-07-07
CORS support fails with credentials & All origin allowing


treerao:
I'm encountering a problem with the CORS support when using credentials.  In particular, I get the following error in latest Chrome, latest Cornice:

> XMLHttpRequest cannot load '....' A wildcard '*' cannot be used in the 'Access-Control- Allow-Origin' header when the credentials flag is true. Origin '....' is therefore not allowed access.

What I think should be happening is the server should be setting the 'Access-Control-Allow-Origin' response header to the origin provided in the credentialed request.

Looking at the source, first I notice that line [cors.py:100](https://github.com/mozilla-services/cornice/blob/master/cornice/cors.py#L100) is checking for a header on the request, Access-Control-Allow-Credentials, which i believe is meant to be a response header.  If instead the test checked the service for credentials support as it does on [cors.py:124](https://github.com/mozilla-services/cornice/blob/master/cornice/cors.py#L124), then the action I would want would happen here as apparently intended.  Alternatively this Access-Control-Allow-Origin can be set right there in the then of that test to the origin.

ametaireau:
Hi!

Thanks for reporting the issue. It seems pretty straightforward to fix. Would you like to work on a fix?

treerao:
Sure, submitted a pull request now.

leplatrem:
As far as I understand, this was merged :)

PR #320, opened 2015-07-08
fix check for cors allow credentials

treerao:
As per my submission of #319, I was hitting  CORS issue when using credentials.  In particular, I get the following error:

> XMLHttpRequest cannot load '....' A wildcard '*' cannot be used in the 'Access-Control- Allow-Origin' header when the credentials flag is true. Origin '....' is therefore not allowed access.

What I think should be happening is the server should be setting the 'Access-Control-Allow-Origin' response header to the origin provided in the credentialed request. I've replaced an incorrect check of a request header, Access-Control-Allow-Credentials, which is really a response header with a correct test on the service definition.

### DIFF
--- a/cornice/cors.py
+++ b/cornice/cors.py
@@ -96,8 +96,7 @@ def ensure_origin(service, request, response=None):
                         for o in service.cors_origins_for(method)]):
                 request.errors.add('header', 'Origin',
                                    '%s not allowed' % origin)
-            elif request.headers.get(
-                    'Access-Control-Allow-Credentials', False):
+            elif service.cors_support_credentials_for(method):
                 response.headers['Access-Control-Allow-Origin'] = origin
             else:
                 if any([o == "*" for o in service.cors_origins_for(method)]):
--- a/cornice/tests/test_cors.py
+++ b/cornice/tests/test_cors.py
@@ -238,6 +238,8 @@ class TestCORS(TestCase):
             })
         self.assertEqual(resp.headers['Access-Control-Allow-Origin'],
                          'lolnet.org')
+        self.assertEqual(resp.headers['Access-Control-Allow-Credentials'],
+                         'true')
 
     def test_responses_include_an_allow_origin_header(self):
         resp = self.app.get('/squirel', headers={'Origin': 'notmyidea.org'})
--- a/cornice/tests/test_cors.py
+++ b/cornice/tests/test_cors.py
@@ -228,15 +228,6 @@ class TestCORS(TestCase):
         self.assertNotIn('Access-Control-Allow-Origin', resp.headers)
         self.assertEqual(resp.json, 'squirels')
 
-    def test_resp_allow_origin_wildcard(self):
-        resp = self.app.options(
-            '/cors_klass',
-            status=200,
-            headers={
-                'Origin': 'lolnet.org',
-                'Access-Control-Request-Method': 'POST'})
-        self.assertEqual(resp.headers['Access-Control-Allow-Origin'], '*')
-
     def test_origin_is_not_wildcard_if_allow_credentials(self):
         resp = self.app.options(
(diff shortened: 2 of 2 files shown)


treerao:
Travis runs of tests generates fails in 2 environments py26 & py34 are unrelated.  The py27 has test fail that is I believe a bad test because it assumes the same thing the original code did i.e. that you can use Access-Control-Allow-Credentials as a request header.  I'm researching this further now.

treerao:
apologies for that broken commit message there.  What I would add is that when eliminate the incorrect use of 'Access-Control-Allow-Credentials' as a request header in that one test, then it and the previous test can not both pass since they are the same and expect different results.  Eliminated the test that is incorrect based on the issue that triggered this fix.

cornice/cors.py
```diff
@@ -96,8 +96,7 @@ def ensure_origin(service, request, response=None):
                         for o in service.cors_origins_for(method)]):
                 request.errors.add('header', 'Origin',
                                    '%s not allowed' % origin)
-            elif request.headers.get(
-                    'Access-Control-Allow-Credentials', False):
+            elif service.cors_support_credentials_for(method):
```
ametaireau: I believe this is not the correct way of handling the problem: here it prevents the response to contain the correct AC-Allow-Origin response header (as the tests showcase).

This is probably due to the fact this is in an elif where the else tests that the allowed origins contain a wildcard.

cornice/tests/test_cors.py
```diff
@@ -228,15 +228,6 @@ def test_resp_dont_include_allow_origin(self):
         self.assertNotIn('Access-Control-Allow-Origin', resp.headers)
         self.assertEqual(resp.json, 'squirels')
 
-    def test_resp_allow_origin_wildcard(self):
```
ametaireau: I don't understand why this test should go away.

cornice/cors.py
treerao: I believe the test is testing for something that in fact should not be true and that's why I eliminated the test.  Let me explain there on your other comment.

cornice/tests/test_cors.py
```diff
@@ -228,15 +228,6 @@ def test_resp_dont_include_allow_origin(self):
...
-        resp = self.app.options(
-            '/cors_klass',
-            status=200,
-            headers={
-                'Origin': 'lolnet.org',
-                'Access-Control-Request-Method': 'POST'})
-        self.assertEqual(resp.headers['Access-Control-Allow-Origin'], '*')
```
treerao: Two ways to support removing this test case:

1) it is technically incorrect in CORS to return what this test wants to see i.e. the * should not be returned in AC-Allow-Origin when withCredentials is used.  I found this explicitly stated in Monsur Hossain's book CORS in Action and also in a few StackOverflow threads, most [notably here](http://stackoverflow.com/questions/19743396/cors-cannot-use-wildcard-in-access-control-allow-origin-when-credentials-flag-i/19744754#19744754)

2) The test right after this eliminated test has the same issue that is resolved in the fix.  The AC-Allow-Credentials header is not a request header, but rather a response header, so it shouldn't be set by the client.  When you delete it, you then have 2 tests that expect contradictory results, so one needs to be eliminated anyway.

Make sense?

ametaireau: Yes! I believe this fix then misses a test that exercise exactly what you described (we should check that the header is returned correctly, but not sent).

ametaireau:
Okay, this makes more sense now. Thanks for waiting this long (I was on vacations). I believe we need to add tests still there and then we'll be good.

treerao:
Hope you enjoyed your vacation.  Added the assertion to the fixed test.  Thanks.

ametaireau:
Thanks! Merged.

PR #327, opened 2015-07-20
added form level validation for colander schemas

winstonf88:
Schema level validators are not called during request validation. 
In the example bellow the ```validate_password_match``` function is never called. 
This pull request fixes that.

```python
from cornice import Service
user_service = Service('user', '/api/users')

def validate_password_match(schema, appstruct):
    if appstruct['password'] != appstruct['password_confirm']:
        error = colander.Invalid(form)
        error['password_confirm'] = 'Passwords do not match'
        raise error

class UserSchema(colander.Schema):
    name = colander.SchemaNode(colander.String())
    password = colander.SchemaNode(colander.String())
    password_confirm = colander.SchemaNode(colander.String())

schema = UserSchema(validator=validate_password_match)

@user_service.post(schema=schema)
def add_user(request)
    ...
    ...

    
```

### DIFF
--- a/cornice/schemas.py
+++ b/cornice/schemas.py
@@ -178,3 +178,19 @@ def validate_colander_schema(schema, request):
         for param in set(params) - set([attr.name for attr in attrs]):
             request.errors.add('body' if param in body else 'querystring',
                                param, msg % param)
+
+    if schema.colander_schema.validator:
+        try:
+            schema.colander_schema.validator(
+                schema.colander_schema,
+                request.validated
+            )
+        except Invalid as errors:
+            for error in errors.children:
+                # import ipdb; ipdb.set_trace()
+                request.errors.add(
+                    getattr(error.node, 'location', 'body'),
+                    error.node.name,
+                    '\n '.join(error.messages())
+                )
+


surabujin:
Will you also emulate scheme level "preparer", "default", "missing" and other colander features?

PS I am working on completely different implementation of colander integration. Current one is terrible!

winstonf88:
I could definitely implement these colander features. I just don't understand what is the real diference between the diferent "location" values (querystring, headers, body, path). Is that part of the REST concept? Or is it more of a cornice concept? 
If you/someone could explain this to me, I would be happy to try to improve these features.

ametaireau:
@surabujin Please, do! I realize that the current state of the code isn't good at all regarding colander, I would happilly review and merge this if appropriate.

ametaireau:
querystring, headers, body, path is the location of where cornice will check for the data.

- querystring is what's in the querystring (e.g after the `?` in the URL)
- headers is http headers
- body is the body of the request (for POST, PUT, DELETE)
- path is what's in the path part of the URL (depending the variables you defined in your URL schemes).

Hope that makes it a little clearer.

cornice/schemas.py
```diff
@@ -178,3 +178,19 @@ def _validate_fields(location, data):
         for param in set(params) - set([attr.name for attr in attrs]):
             request.errors.add('body' if param in body else 'querystring',
                                param, msg % param)
+
+    if schema.colander_schema.validator:
```
ametaireau: I wasn't aware about this way of doing validation for colander, nice.

ametaireau:
In order to be merged, this pull request also need to be exercised through the test suite. Please, ask me if you need help to understand how to create a test!

winstonf88:
@ametaireau Thanks for the explanation. I will come up with a better solution on this and maintain all colander scheme level features. As soon as its done I'll submit it (with tests).

tisdall:
There's discussion on colander to remove arbitrary keywords on the `__init__` and that change would effectively break the `location` and `type` keywords being used on the colander schemas.  I've yet to really understand the point in the `type` keyword over just using proper colander types so I don't think removing that is really that big a deal (until I figure out why it was added in the first place).  The `location` functionality could be replaced by making all validation schemas look something like:

```
class ValidationSchema(MappingSchema):
    querystring = SchemaNode()
    headers = SchemaNode()
    body = SchemaNode()
    path = SchemaNode()
```

So, each of the 4 "locations" would be represented by children of the validation schema.

That's the change I'd like to see done.

Issue #190, opened 2013-10-22
Cornice skips the `preparer` and `validator` steps of a SchemaNode subclass


abourget:
Having something like this:

    class TransactionsPostSchema(SchemaNode):
        def preparer(self, value):
            value = merge_src_token_with_src_account(value)
            return value

        def validator(self, node, value):
            .. do stuff ..

        src_account = SchemaNode(String(), missing=None)
        src_token = SchemaNode(String(), missing=None)
        dst_account = SchemaNode(String())

When using with:

    @my_service.post(schema=TransactionsPostSchema)
    def my_service_post(request):
        ....

Cornice disregards the `preparer` and `validator` at this level, because it loops on the `.children` of the root node only.  It would need to either re-apply manually the calls to the `preparer` (that we find in colander's `_SchemaNode::deserialize` function) and `validator`.  That wouldn't be very DRY though.

The other option might be to instead call .deserialize() on the root Schema after checking that the attributes are "present" at the correct location. This way, colander would do the job it does normally.

So what I understand, is that the passed-in Schema is only used as a "container" for its children.  In that case, how would I achieve what I'm trying to do here, without passing "validators=my_prepare_and_validators" on each service view config ?

As a side note, shouldn't it be documented that the `location` element is only verified on the first level ? On the immediate children of the root schema passed in the view configurator.  It wasn't clear to me where the `location` should be, or even if it was required at all.

alexsv:
+1

delijati:
+1

leplatrem:
This will be solved in Cornice 2.0 (c.f. current master)

PR #329, opened 2015-07-31
Patch to resolve issue #190.

antigenius:
Currently cornice iterates over colander schemas and deserializes each
node individually. I've refactored to allow colander to handle the
work, instead deserializing at the Schema rather than SchemaNodes.

There was also some confusion in the test regarding colander.default
and colander.missing. `default` is used on serialization while
`missing` is used on deserialization, which is what we want.

### DIFF
--- a/cornice/schemas.py
+++ b/cornice/schemas.py
@@ -107,7 +107,7 @@ def validate_colander_schema(schema, request):
         raise SchemaError('colander schema type is not a Mapping: %s' %
                           type(schema_type))
 
-    def _validate_fields(location, data):
+    def _extract_fields(location, data):
         if location == 'body':
             try:
                 original = data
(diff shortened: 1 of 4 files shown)


tisdall:
`deserialize(serialize())` ... that's the gist of validation via colander.  So, you do need to take into account both `default` and `missing`.  Though, you're new code seems to be going a different route...

The issue with colander is it's not intended for dealing with JSON.  Colander has a serialized format called "cstruct" that's dicts with all strings.  The `deserialize(serialize())` method is how to shoe-horn json validation by using colander.  If colander is ever changed to use json as it's serialized format than maybe you could do just `deserialize`, but until that happens `deserialize` alone just won't cut it.

antigenius:
My understanding was that by the time the colander schema is called, any JSON supplied in the request had itself already been deserialized and attached to the request object. All that needed to be done was parsing out the data values based on the `location` keyword allowing colander to work it's magic there.

tisdall:
@antigenius - yes, that's correct.  However the json deserialized does not equal a cstruct.

Let's say you had '{"x":3}' as the body.  The `simplejson` will deserialize that into `{'x': 3}` which is not a valid cstruct.  Colander's cstruct has only strings so calling colander's deserialize on `{'x': 3}` will probably choke.  If you serialize `{'x': 3}` with colander you get `{'x': '3'}` which is the cstruct format.

antigenius:
Ah. So, I wasn't having issues with colander deserializing values other than strings. It does support it, though it appears that's [not supported behavior](https://github.com/Pylons/colander/issues/80). Is there any danger in simply wrapping the [cstruct assignments](https://github.com/antigenius/cornice/blob/master/cornice/schemas.py#L135-L137) in str()? That would ensure the cstruct supplied to deserialize conforms to what colander is expecting.

tisdall:
@antigenius - you could go through and wrap values in str(), but why?  When you call colander's serialize it cycles through all the elements and effectively calls str() on everything for you.  Additionally, it will call preparers on the values as there may be something else that needs to happen than just calling str() on it.

Please take a look at #330 as I'm trying to solidify the things I've been talking about into a PR.

antigenius:
I wanted to avoid the serialize since we just need to deserialize it again anyway. Seemed like an unnecessary step and since the code was already iterating over the nodes, it made sense to stringify things there. As preparers are called on both deserialization and serialization, I figured it was a "two birds, one stone" scenario.

Some background on why I submitted this PR. I was trying to run a validator at the schema level to compare to nodes to each other (email and email_confirmation). The way cornice handles it now, it skips schema level preparers and validators.

bbangert:
@tisdall I would think that raw values from a request should not be treated the same as serializing a colander instance, ie, its preparers *should not be run*. The point of validation and deserialization is the assumption that it was never serialized to begin with (ie, its validating/deserializing raw data). 

Because colander was designed with form params as the target, the assumption (and design) is such that the values are all strings, because with form/query params, thats all you get. So, if the body is JSON, and then validation is desired, the best way to get it to a form/query-param like state would be to recursively str every value that isn't already a str. It's not ideal, but it gets the input to a similar state with the least performance hit. Running serialize first, then deserialize is not a performance hit I'd want to take, and it needlessly runs over the headers/path which are already string values ready for deserialize.

Is there something I'm missing?

tisdall:
@antigenius You're right, the "deserialize(serialize())" isn't the most efficient, but I think it's cleanest way to make colander do this kind of validation (which it was never intended to do, really).  It also makes sure all the steps are done, nothing is left out, and it's way easier to test.  The current methods in Cornice are trying to re-implement components of colander but are missing things.  It's also adding a bunch of unnecessary things like an additional "type" and a "location" component (which could be handled as part of a Mapping).

tisdall:
@bbangert - good points...  I was kind of relying on #190 as being a reason that we should support preparers without really thinking of a good use-case for having that.  Also, I was going towards the "deserialize(serialize())" because Cornice is actually already doing that in some places (https://github.com/mozilla-services/cornice/blob/master/cornice/schemas.py#L137) but just not consistently.

However, iterating over everything to apply `str()` to it is pretty much the same thing as calling colander's `serialize()` except that it will ignore `default` and `preparer`.  I don't think you can improve speed by testing if the value is already a string over just calling str() on everything (I haven't tested that assumption, though).

I was also leaning towards just letting colander handle as much as possible as it's already well tested.

tisdall:
@bbangert - I've been using "deserialize(serialize())" with my own validation and one issue that comes up is when you nest Mappings.  If you want a portion of the Mapping to be optional you need to add both `default=colander.drop` and `missing=colander.drop` or it will raise an exception.  If you skip the serialization step and just do a recursive str() then that simplifies the schema quite a bit.  We'd then just drop looking at 'default' and use just 'missing'.  Unless someone sees a need to keep 'default' and preparers then this is probably a better way to go.

antigenius:
@tisdall Preparers are only called during deserialization. As such, you'd only need 'missing'.

cornice/schemas.py
```diff
@@ -123,51 +123,61 @@ def _validate_fields(location, data):
...
+                     ('body', body), ('querystring', qs)]:
+        meta = _extract_fields(location[0], location[1])
+        cstruct.update(meta['cstruct'])
+        attr_locs.update(meta['locations'])
+
+    try:
+        appstruct = schema.colander_schema.deserialize(cstruct)
+    except Invalid as e:
+        for k, v in e.asdict().items():
+            v = '%s is missing' % k if v == 'Required' else v
```
tisdall: colander has a localization component.  If it's used than it'll translate "Required" into one of the other languages and this match won't work any more.

tisdall:
@antigenius - oops... yeah, so only 'default' would be ignored by doing the str() thing.

antigenius:
@tisdall Correct, but only if you're serializing. cornice docs state that the `schema=` keyword is strictly for deserialization, so I think ignoring 'default' is implicitly okay. But assumptions being what they are...

tisdall:
@antigenius - well the tests and schema.py are explicitly looking at `default` right now...  I don't think there's much of a performance boost of doing str() in Cornice, but I think ignoring `default` may make more sense overall and make it worth doing.  As I pointed out about my issue with nested Mappings... it'd be nice to drop the default clauses in my validation schemas.

tisdall:
@antigenius - I've updated #330 to manually create the cstruct instead of using `serialize`.

antigenius:
Excellent! I'll check it out.

leplatrem:

Issue #310, opened 2015-06-15
addition of `type` attribute on colander elements


tisdall:
Colander elements already have colander types, but the examples in Cornice and the tests seem to have added another attribute called `type`.  The code indicates that if you don't add this new `type` it just uses the colander `typ` instead.  What is the purpose of this new `type` attribute?

delijati:
+1

leplatrem:
This was fixed in Cornice 2.0 (current master)

Issue #249, opened 2014-09-18
Schema validation may throw an uncaught exception with valid JSON


ltvolks:
A service validating with a colander schema requires that the request body deserialize to a dictionary. 

If the JSON body deserializes to anything else, an uncaught exception will be thrown. The example below demonstrates the behavior. There we attempt to post JSON with a top-level Array (valid per spec), and a colander exception bubbles up.

It doesn't matter what type of validation schema is defined since the cornice behaviors [here](https://github.com/mozilla-services/cornice/blob/master/cornice/schemas.py#L98) and [here](https://github.com/mozilla-services/cornice/blob/master/cornice/schemas.py#L150) rely on the schema being a `colander.Mapping` type. (I thought a `colander.SequenceSchema` should work as a validator but that is another issue),

```python
import colander
from cornice.service import Service
from pyramid.config import Configurator
from webtest import TestApp


class ValidationSchema(colander.MappingSchema):
    key = colander.SchemaNode(colander.String())

svc = Service(name="test", path="/test")
svc_validation = Service(name="test-validation", path="/test-validation")


@svc.post()
def post(request):
    return {"test": "succeeded"}


@svc_validation.post(schema=ValidationSchema)
def post(request):
    return {"test": "succeeded"}


config = Configurator()
config.include("cornice")
config.scan()
app = config.make_wsgi_app()

client = TestApp(app)

# no validation performed...the deserialized list is available to the view
client.post_json('/test', [1, 2, 3])

# Throws an exception
client.post_json('/test-validation', [1, 2, 3])
```

Traceback:
```python
  File "/venvs/cornice-test/src/cornice/cornice/service.py", line 495, in wrapper
    validate_colander_schema(args['schema'], request)
  File "/venvs/cornice-test/src/cornice/cornice/schemas.py", line 147, in validate_colander_schema
    _validate_fields('body', body)
  File "/venvs/cornice-test/src/cornice/cornice/schemas.py", line 99, in _validate_fields
    data = webob.multidict.MultiDict(schema.unflatten(data))
  File "/venvs/cornice-test/src/cornice/cornice/schemas.py", line 81, in unflatten
    return self.colander_schema.unflatten(data)
  File "/venvs/cornice-test/lib/python2.7/site-packages/colander/__init__.py", line 1880, in unflatten
    paths = sorted(fstruct.keys())
AttributeError: 'list' object has no attribute 'keys'
```

[...]

Issue #211, opened 2013-12-28
need a way to trap unknown keys in 'querystring' of colander schema


joesteeve:
Cornice silently ignores keys that are not found in a colander schema. This is not favourable with the 'querystring' location. Unknown keys in the 'querystring' is an error. There should be a way to specify this in the schema so that cornice can report the error to the client.

leplatrem:
With the new approach in Cornice 2.0 (current master), this will be possible via Colander (cf. `unknown='raise'`: http://docs.pylonsproject.org/projects/colander/en/latest/api.html?highlight=missing#types)

Issue #173, opened 2013-07-25
support for flattened schema in querystring


matleh:
The API I am workin on needs query-parameters like `?x.a=123&x.b=456`. Colander has the unflatten method for these use-cases, but I can not see, how I can make use of that when using Cornice.

It looks like `validate_colander_schema` needs code to explicitly support that use-case. I could try to work on that, if there are no other existing ways to achieve that - maybe using some clever colander tricks, like preparer or defered-fields.

Did anybody have a similiar use-case before? What are your opinions on this?

leplatrem:
This should be fixed with the new API in Cornice 2.0 (current master) :)

PR #330, opened 2015-07-31
reworking colander schema validation [WIP]

tisdall:
Okay, due to all the issues and comments surrounding colander validation I started working on a replacement.  Basically a lot of things were being done in Cornice when they could have easily been left to Colander (like iterating over child elements).  The schema.py file is now 62 lines where the old one was 180.  The 4 main components (querystring, headers, body, and path) are grouped into one structure and passed through a single schema for validation tests.  Each of those 4 components, if validated, are moved to `request.validated`.  Any errors are put in `request.errors`.

This is obviously a huge change from the previous method, so I wanted to get some input before I complete all aspects of this PR (please ignore the failing tests for the moment).

I modified a few existing tests to kind of demonstrate how the system works on validating the body but obviously more are needed to test validating the other 3 types.

This deals with the following issues: #310, #211, #190, #327, #193, #249 

Perhaps we could support #173 by calling unflatten in the validator.

### DIFF
--- a/cornice/schemas.py
+++ b/cornice/schemas.py
@@ -16,18 +16,34 @@ else:
             return colander.Mapping(unknown='raise')
 
     class CorniceSchema(colander.MappingSchema):
-        querystring = StrictMappingSchema(
-            default=colander.drop
-        )
-        headers = StrictMappingSchema(
-            default=colander.drop
-        )
-        body = StrictMappingSchema(
-            default=colander.drop
-        )
-        path = StrictMappingSchema(
-            default=colander.drop
-        )
+        querystring = StrictMappingSchema()
+        headers = StrictMappingSchema()
+        body = StrictMappingSchema()
+        path = StrictMappingSchema()
+
+    def simple_cstruct_serialize(val):
+        """ cstruct is colander's internal use format which is a
+        series of nested dicts, lists, and tuples with only str values
+        as the "leaf" values.  If this method will try to recursively
+        go through a given value and convert all dict-like objects into
+        dicts, convert iterables that aren't dict-like into lists, and
+        convert everything else into a string.
+        """
+        try:
+            # try dict-like interpretation
+            result = {}
+            for k in val.keys():
+                result[k] = simple_cstruct_serialize(val[k])
+            return result
+        except (TypeError, AttributeError):
+            try:
+                # try iterable interpretation
+                result = []
+                for k in val:
+                    result.append(simple_cstruct_serialize(k))
+                return result
+            except TypeError:
+                return str(val)
 
 
 def validate_colander_schema(schema, request):
@@ -50,7 +66,7 @@ def validate_colander_schema(schema, request):
     }
 
     try:
-        cstruct = schema.serialize(initial_appstruct)
+        cstruct = simple_cstruct_serialize(initial_appstruct)
         appstruct = schema.deserialize(cstruct)
     except colander.Invalid as e:
         for component_path, msg in e.asdict().items():
--- a/cornice/schemas.py
+++ b/cornice/schemas.py
@@ -1,180 +1,62 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
-import webob.multidict
-
-from pyramid.path import DottedNameResolver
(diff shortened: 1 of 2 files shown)


tisdall:
BTW, don't bother looking at the diff's for the two files as they're pretty drastic changes and it's probably easier to just read the whole file...

tisdall:
As per the discussion on #329, I've updated this PR to not call colander's `serialize()` any more and just manually create the cstruct.  This way the validation schema only has to take the `deserialize()` process into account.

tisdall:
I'd like to hear back from someone with the ability to merge this PR before continuing work on it...  I don't want to spend more time on it if it won't be used.

leplatrem:
Hi @tisdall,

Thanks a lot for tackling this! It was extremely necessary!

You can definitely go on with making the tests pass :) I'll take some time to review the pull-requests about schemas (like #337), and give as much feedback as I can :)

Thanks (*again*) :)

cornice/schemas.py
```diff
@@ -1,180 +1,78 @@
...
+            result = {}
+            for k in val.keys():
+                result[k] = simple_cstruct_serialize(val[k])
+            return result
+        except (TypeError, AttributeError):
+            try:
+                # try iterable interpretation
+                result = []
+                for k in val:
+                    result.append(simple_cstruct_serialize(k))
```
leplatrem: When looking at this: https://travis-ci.org/mozilla-services/cornice/jobs/76151945#L525 recursivity seems infinite!

cornice/schemas.py
```diff
@@ -1,180 +1,78 @@
...
+    class StrictMappingSchema(colander.MappingSchema):
+        @staticmethod
+        def schema_type():
+            return colander.Mapping(unknown='raise')
+
+    class CorniceSchema(colander.MappingSchema):
+        querystring = StrictMappingSchema()
+        headers = StrictMappingSchema()
+        body = StrictMappingSchema()
+        path = StrictMappingSchema()
```
leplatrem: For retro-compatibility, we could add a ``from_colander`` method (?)

```python
    @classmethod
    def from_colander(klass, colander_schema, **kwargs):
        # XXX: ...written on a hungry stomache:
        class BodySchema(CorniceSchema):
            body = colander_schema
        return BodySchema(**kwargs)
```

Thoughts ?

cornice/schemas.py
tisdall: yep.  it seems I didn't properly test `simple_cstruct_serialize()`.  It's because strings are iterable.  I'll patch this.

cornice/schemas.py
tisdall: This PR works totally different then the past system so maybe it'd be best to actually change the name so there's no longer a CorniceSchema.  Or, CorniceSchema is some sort of compatibility class that's deprecated that helps people transition to the new system.

Any ideas for a different name?  Maybe it should be "ColanderSchema" since we're talking about adding alternate schema libraries and this class is already within the `cornice` namespace?

tisdall: Is "CorniceColanderSchema" too wordy?

tisdall: Another issue with the old system is everything ends up in `request.validated` so, for example, if you have a "body" element and a "querystring" with the same name they'll conflict.  The new one ends up with `request.validated['body']` and `request.validated['querystring']`.  I wrote a `from_colander()` method, but the result in request.validated will be completely different then before.

I don't think there's a way to support a smooth transition from the old...  So, I think it's best to fail early and hard so people are forced to face the change directly.

delijati:
This is perfect. I'm currently working on #285 swagger integration and it would make the whole process of generating a **consistent** swagger schema much easier. Ony one schema for ever location. So only for body we could allow nested schemes.

The current validation process is too complicated a simple `schema.deserialize(data)` would be enough, if we got rid of the `location` and use this pull request (seperate schema for body, path, querystring, header).

delijati:
This would make it possible to use https://github.com/stefanofontanelli/ColanderAlchemy ;)

tisdall:
Hey guys...  I'm not really working on the project where I used cornice any more and other responsibilities  have come up that make it even more unlikely that I'll be able to finish this.  Hopefully someone else can pickup where I left off.

leplatrem:
See how we resolved this https://github.com/Cornices/cornice/blob/master/docs/source/upgrading.rst#colander-validation

delijati:
Nicely solved :)

Issue #334, opened 2015-09-09
add_view; includeme not working as expected


delijati:
Following a minimal example of a app structured with `includeme` and `add_view`. I marked the parts with `FIXME` that are not working like expected (thats how pyramid also allows them).

*demoapp/__init__.py*:
````python
from pyramid.config import Configurator

def main(global_config, **settings):
    config = Configurator(settings=settings)
    config.include("cornice")
    config.include("demoapp.views.includeme")
    # FIXME shouldn't be necessary -> config.scan("demoapp.views")
    return config.make_wsgi_app()
````

*demoapp/views.py*:
```python
from collections import defaultdict
from pyramid.view import view_config
from cornice import Service

_USERS = defaultdict(dict)

# FIXME this should also work in includeme
user_info = Service(name='users', path='/{username}/info')

def get_info(request):
    username = request.matchdict['username']
    return _USERS[username]

def set_info(request):
    username = request.matchdict['username']
    _USERS[username] = request.json_body
    return {'success': True}

def who(request):
    return {"usernames": _USERS.keys()}

def includeme(config):
    # FIXME user_info should be defined here
    user_info.add_view('post', set_info)
    user_info.add_view('get', get_info)

    config.add_route("who", "/who")
    config.add_view(who, route_name="who", renderer="json")
````

The easiest solution would be, to add `config` as a paramater to Service!?

ametaireau:
`Service` doesn't currently have any notion of pyramid `config`. We do all the registration in pyramid in `pyramidhook` instead.

In the second case, I don't get why `user_info` (that you defined) isn't present where you put the FIXME.

Also, I don't quite get what you're trying to do here, as it seems you don't want to use the `Service` `add_view` directly. Cornice provides you decorators to ease this process, actually :-)

delijati:
If i put `user_info` in `def includeme`, the views are not registered. 

I like the pyramid *imperative* style of defining views and grouping all needed information in one place `def includeme` (separating views and configuration).

Mixing *declarative*  and *imperative* with `@view_config` and `add_route` and in worst case seperating them in two files, makes it for me to cluttered.

Anyway i got it working by calling `add_cornice_service` see `XXX`:

```python
from collections import defaultdict
from cornice import Service

_USERS = defaultdict(dict)

def get_info(request):
    username = request.matchdict['username']
    return _USERS[username]

def set_info(request):
    username = request.matchdict['username']
    _USERS[username] = request.json_body
    return {'success': True}

[...]

PR #335, opened 2015-09-09
adding abillity to define services imperative

delijati:
See issue #334 for more info. Coverage for sphinxext increased to 60% :)

### DIFF
--- a/cornice/resource.py
+++ b/cornice/resource.py
@@ -11,7 +11,7 @@ except ImportError:
     VENUSIAN = False
 
 
-def resource(depth=1, **kw):
+def resource(depth=2, **kw):
     """Class decorator to declare resources.
 
     All the methods of this class named by the name of HTTP resources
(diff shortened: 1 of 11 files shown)


cornice/resource.py
```diff
@@ -24,83 +24,127 @@ def resource(depth=1, **kw):
...
+    All the methods of this class named by the name of HTTP resources
+    will be used as such. You can also prefix them by "collection_" and they
+    will be treated as HTTP methods for the given collection path
+    (collection_path), if any.
+
+    Here is an example::
+        class User(object):
+            pass
+
+        add_resource(User, collection_path='/users', path='/users/{id}')
```
ametaireau: If possible, add the `:params:` syntax which explains what is contained in the arguments.

docs/source/resources.rst
```diff
@@ -30,6 +30,24 @@ Here is how you can register a resource::
             _USERS[len(_USERS) + 1] = self.request.json_body
             return True
 
+Here a example of imperative way defining a resource::
```
ametaireau: consider: "Here is an example of how to define cornice resources in an imperative way"

docs/source/resources.rst
```diff
@@ -30,6 +30,24 @@ Here is how you can register a resource::
...
+
+    from cornice.resource import add_resource, add_view
```
ametaireau: I would rather import `resource` and use `resource.add_view` and `resource.add_resource`.

ametaireau:
Hi!

Thanks for taking the time to submit this pull request. That's appreciated. I like how the decorators are now completely optional, and in some way, it closes the gap I started to fill when I did [the cornice refactor](http://blog.notmyidea.org/refactoring-cornice.html).

I'm not a big fan of the `add_resource` and `add_view` names, but they seem to match with the ones provided by pyramid itself, so I think it's *fine* to let it like that.

I've added a few comments and I think we're good for merging once they're addressed.

delijati:
Hi,

i've added the directive `add_cornice_resource` to get rid of the `scan`. There is still that issue with ordering of the resource views. I think a cleaner way would be to add a `Register` Object like the Service object witch holds the collected informations. But this should be done in another refactoring ;)

delijati:
Do you had some time to look into that?

delijati:
@leplatrem could you also take a look into that pull request

ametaireau:
Hey Josip, this looks really good. Eveything is in there: impelmentation, docs, new tests, plus the old tests are passing.

Let's merge :-)

delijati:
Wonderful do you know if someone is already working on the push request https://github.com/mozilla-services/cornice/pull/285 ? If not i will look into that one next week.

ametaireau:
Neat. Nobody is currently working on this I believe. Feel free to take it!

Issue #274, opened 2015-02-16
marshmallow (de)serialization


philtay:
Are you interested in adding support for [marshmallow](https://github.com/marshmallow-code/marshmallow) in cornice? marshmallow is a (de)serialization library and I think it would be really good to have it integrated in cornice (alongside or even instead of colander). I can open a PR for this if you wish.

ametaireau:
I would be happy to have something in cornice that let you use whatever deserialisation mechanism.

I believe it's already the case and it would be pretty easy to add validation logic that uses marshmallow rather than colander. If it's not the case, we should of course make cornice evolve to support that.

leplatrem:
With cornice 2.0 (current master), it'll be very easy to plug marshmallow (cf. #397)

PR #206, opened 2013-12-13
Allow i18n of colander error messages

amarandon:
Fix #200

After investigating the issue further, I realized that not everything can be done by Colander. Colander can only provide translation strings but it cannot perform the translation itself because it doesn't depend on Pyramid and doesn't know about localizers. As suggested earlier, I've proposed a change to Colander so that we can pass a translation callable to the ``asdict`` method: https://github.com/Pylons/colander/pull/157

However it's still up to Cornice to provide a callback that is aware of the current locale.
Since Cornice aims at following HTTP conventions, I've added support for the HTTP ``Accept-Language`` header. I've also added some configuration code that sets up a localizer depending on configuration variables based on suggestions from the Pyramid documentation:
 * [``pyramid.default_locale_name``](http://docs.pylonsproject.org/projects/pyramid/en/1.3-branch/narr/i18n.html#localization-related-deployment-settings) to specify the default locale
 * [``available_languages``](http://docs.pylonsproject.org/projects/pyramid/en/1.3-branch/narr/i18n.html#detecting-available-languages) to specify the list of available languages

PR #337, opened 2015-09-14
Rewrite colander schema

surabujin:
I am propose modular interface to schema validation. It's interface and behaviour mostly compatible with existing schema validator.

But it use "clean" colander deserialisation call, so all existing hacks that emulate in incompatible way colander features are removed.

PS New interface allow add other schema validators.


surabujin:
Depends from https://github.com/Pylons/colander/pull/241

leplatrem:
Thanks a lot @surabujin!

I really like the fact to be able to validate the data with any library! It covers issue #274 (and stuff like https://github.com/mozilla-services/cliquet/issues/244). Plus it looks like you managed to write some backward compat stuff. Neat!

The diff is pretty huge to review,  did you write specific tests for the new code?

What do you think if you split the Colander usage refactor from the pluggability (based on #330)? Is that even possible?

Thanks again for this huge piece of work!

surabujin:
I have not add any specific tests yet. I just reach first goal - make it pluggable, leave colander work to colander and make it as much as possible to previous API (it can be even more compatible, but I don't see a sense for it now).

Even if I can split this 2 change - make schemas pluggable and rewrite colander support, I can't make to work old colander validation code with proposed pluggable model. So one patch can't be used without another one.

tisdall:
I like the idea of making the schema checking pluggable as there's been other requests for using other validation libraries already.  I'm having a little trouble understanding how this patch works, though...  Do you have any example code that you used to test when developing?

surabujin:
@tisdall you want examples how other libraries can be integrated here? I don't have them yet.

I will update documentation, also I have a plan to add support of FormEncode.

tisdall:
No, I was thinking more along the lines of examples of how this works with
existing colander schemas.

surabujin:
Promised usage example: https://gist.github.com/surabujin/8109f55dd02b57c17fda

leplatrem:
> Promised usage example

I like it, even though I wonder if we shouldn't get rid of those custom `location` parameters, as in #330 

> Depends from Pylons/colander#241

Unfortunately, the merge and release cycles of colander are very long. See for example the PR #206, which depends on https://github.com/Pylons/colander/pull/157. The PR was merged more than a year ago in colander, but never got released.

That's also why having pluggable schema libraries is a must-have! Projects like https://github.com/nicolaiarocci/cerberus or https://github.com/marshmallow-code/marshmallow are a lot more active and reactive!

Nevertheless, please be aware that we cannot merge so much new code without exhaustive and «specs» tests [1]. Especially because afterwards, we become the implicit maintainers of all this :)

Good luck and thank again for this excellent initiative :)

[1] : http://blog.mathieu-leplatre.info/your-tests-as-your-specs.html <br/> https://medium.com/javascript-scene/what-every-unit-test-needs-f6cd34d9836d

lmctv:
While still not in a public release, Pylons/colander#241 have been merged

leplatrem:

PR #338, opened 2015-09-25
Fix Access-Control-Max-Age value if undefined on service

leplatrem:
Otherwise the header is always present a `False` value.

### DIFF
--- a/cornice/service.py
+++ b/cornice/service.py
@@ -508,11 +508,15 @@ class Service(object):
         return False
 
     def cors_max_age_for(self, method=None):
+        max_age = None
         for meth, view, args in self.definitions:
             if method and meth.upper() == method.upper():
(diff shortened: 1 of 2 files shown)


Natim:
r+

cornice/service.py
```diff
@@ -508,11 +508,14 @@ def cors_support_credentials_for(self, method=None):
         return False
 
     def cors_max_age_for(self, method=None):
+        max_age = None
         for meth, view, args in self.definitions:
             if method and meth.upper() == method.upper():
-                return args.get('cors_max_age', False)
+                max_age = args.get('cors_max_age', None)
```
ametaireau: we should break here to avoid continue the for loop. Only the first value matters.

leplatrem: done.

PR #340, opened 2015-10-16
Swagger integration parts of #285

delijati:
Hi i finished my part on the swagger integration. I also included a new scaffold to demonstrate how to get swagger up and running.

This is a pretty big push request and there are still some parts that should be tackled after merging:

1.) The validation process is too complicated, this could be fixed by merging https://github.com/mozilla-services/cornice/pull/330
2.) The sphinx documentation is not extracting nested schemas
3.) There is no validation if someone is setting a different location in a nested schemas, or if someone is using a obj for non body location. (see **brrr**)

```python 
class Person(colander.Schema):
    name = colander.SchemaNode(colander.String(), location="querystring")  # <-- brrr 
    age = colander.SchemaNode(colander.Integer(),
                              validator=colander.Range(0, 200))

class People(colander.SequenceSchema):
    person = Person(location="querystring")  # <-- brrr 

class Schema(colander.Schema):
    ort = colander.SchemaNode(colander.String(), required=False,
                              location="header") # <-- brrr
    people = People(validator=colander.Length(min=1, max=4))
```
4.) Because of 1.) we cant use third party tools like ColanderAlchemy
5.) The code is full of `hasattr(foo, "type")`, why not just depend on `colander`?

I would propose to create a 2.0 backward incompatible version. Colander should be a must dependency for creating validation. This would give a nicer eco system.


leplatrem:
Waho thanks for this huge amount of work!

The scaffold part is pretty heavy. I wonder if that should live elsewhere. @almet what do you think?

My main remark is about the tests. I find them a bit succinct compared to the amount of new code introduced. But we could probably live with that since it's not a critical feature...

almet:
+1 for putting the scaffolding somewhere else (as a separate package maybe).

cornice/ext/swagger.py
```diff
@@ -0,0 +1,207 @@
...
+        tag_name = service.path.split("/")[1]
+        service_tag = dict(name=tag_name)
+
+        path = {"parameters": []}
+
+        # Get path parameters from looking at the path
+        service_path = service.path
+        parameter = dict()
+
+        # Do we have parameters in the path?
```
hamx0r: Lines 117-128 here should be deleted as they

1. duplicate lines 158-168
2. Introduce a bug where the autodetected path parameters in this block are not "updated" in future code (ie Colander SchemaNode), causing duplicate Parameters to show up in Swagger JSON/UI

cornice/ext/swagger.py
```diff
@@ -0,0 +1,207 @@
...
+                if paramType == "querystring":
+                    if service and "{%s}" % name in service.path:
+                        paramType = "path"
+                    else:
+                        paramType = "query"
+                parameter["in"] = paramType
+                parameter["name"] = name
+                if hasattr(attr, "description"):
+                    parameter["description"] = getattr(attr, "description")
+                parameter["required"] = getattr(attr, "required")
```
hamx0r: This line should be replaced with the following two due to how Colander calculates  "required" with respect to `deferred` Schema Binding.  

        missing = getattr(attr, 'missing')
        parameter['required'] = False if (missing == drop or isinstance(missing, deferred)) else True

[Per the source](https://github.com/Pylons/colander/blob/master/colander/__init__.py#L1983), Colander treats `deferred`  attributes as `required`.  I think the logic should be that if a user defers determining what a `missing` SchemaNode/Parameter value should be, then it is clearly not required as some logic is in place to handle it being missing.

elbart:
Hi, what is the current status of this PR? Would be interested in that as well.

leplatrem:
We are very interested by this feature! Ideally, the Swagger approach (now https://openapis.org) would replace the Sphinx ext. 

But we are in the process of refactoring the way `schemas` are managed in Cornice. @tarekziade is working on it \o/

Also, the reality is that we probably won't have enough resource to maintain so much new code, and we might have to split that into a separate package. But first, let's make the schemas API clean, well introspectable and documented !

delijati:
Yes the `schemas` part is pretty messy. If that is cleaned up i will try to get some time extract the swagger part but this a pretty low priority as i'm not using cornice currently. @tarekziade out of curiosity how you gone refactor the `schemas` part? The only ways for me, when i was working on this patch seemed to be making `colander` a dependency.

leplatrem:
@delijati It is not done yet, but want to take the Swagger use-case into account https://github.com/mozilla-services/cornice/pull/376#issuecomment-241998093

alexandre-figura:
Hi there,

I recently discovered the [Core API](http://www.coreapi.org/) project, and I was wondering if you were aware of it?

It provides a generic document to represent a schema or hypermedia response. From this document, you can for example generate a schema in **Open API/Swagger** or **JSON Hyper-Schema** formats.

An implementation example is available on [Django-REST Framework](https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/schemas.py).

Could be interesting to let Core API handle generation of schemas in different formats :+1:

leplatrem:
@delijati We are getting rid of extensions in the Cornice package. I created the Cornices/cornice.ext.swagger repo and added you as a collaborator with write access. You should be able to move your code there, and manage this feature without going through our long and slow review process :)
I'm closing this here for now, don't hesitate if you have question or need help!

delijati:
Is there a api or some docs on how to integrate a extension?

leplatrem:
> Is there a api or some docs on how to integrate a extension?

For this particular PR, it should not be fundamentally different that what you have here. Except that the package will have to be installed separately `pip install cornice_swagger`, and possibly have to be included `config.include("cornice_swagger")`.
Your changes on the template could now go the cookiecutter-cornice package.

Don't hesitate to ask more questions along the way! But basically you can manage the repo as you wish without depending on our slow review process :) We can still link from the cornice docs to promote it

delijati:
Updated: ``https://github.com/Cornices/cornice.ext.swagger`` and ``https://github.com/delijati/cookiecutter-cornice_swagger``

PR #341, opened 2015-10-23
String validators rendered as str.__doc__ in sphinxext

squidsoup:
Currently in sphinxext, validators defined as strings in a resource or service are rendered directly as `str.__doc__` e.g.

    user_info = Service(name='users', path='/{username}/info',
                        validators=('validate'))

will render:

    "str(object='') -> string\n\nReturn a nice string representation of the object.\nIf the argument is a string, the return value is the same object."

This branch provides a fix and a relevant test, type checking strings and ensuring the validator object's `__doc__` is rendered. String to object resolution, also required to render views, has been factored out to a new method, `_resolve_obj_to_docstring`.

This branch also contains some mostly cosmetic refactoring of the sphinext tests.

### DIFF
--- a/cornice/ext/sphinxext.py
+++ b/cornice/ext/sphinxext.py
@@ -121,7 +121,6 @@ class ServiceDirective(Directive):
             service_node += rst2node(trim(service.description))
 
         for method, view, args in service.definitions:
-
             if method == 'HEAD':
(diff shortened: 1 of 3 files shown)


leplatrem:
Looks good to me, thanks!

PR #342, opened 2015-10-30
Convert None to null before calling deserialize

tsauerwein:
I have a schema like the following:

    class TestNoneSchema(MappingSchema):
        foo = SchemaNode(String())
        bar = SchemaNode(Sequence(), SchemaNode(String()), missing=None)

When I try to validate a JSON body where `bar` is `null` (for example `'{"foo": "abc", "bar": null}'`), the validation returns the following error:

     [{'name': 'bar', 'location': 'body', 'description': u'"None" is not iterable'}]

This is because Cornice calls `deserialize` in [validate_colander_schema](https://github.com/mozilla-services/cornice/blob/c71b5c4fb25c70a62b203c9480810ec8707878bf/cornice/schemas.py#L146) with `None` instead of `colander.null`. Because [Sequence.serialize](https://github.com/Pylons/colander/blob/master/colander/__init__.py#L1058-L1079) only checks for `colander.null`, an exception is thrown.

As far as I understand how Colander works, `deserialize` should in general not be called with `None` but `colander.null`. So, it would be the responsibility of Cornice to convert `None` to `null`.

### DIFF
--- a/cornice/schemas.py
+++ b/cornice/schemas.py
@@ -143,6 +143,8 @@ def validate_colander_schema(schema, request):
                             serialized = original.getall(attr.name)
                         else:
                             serialized = data[attr.name]
+                        if serialized is None:
+                            serialized = null
                         deserialized = attr.deserialize(serialized)
                 except Invalid as e:
                     # the struct is invalid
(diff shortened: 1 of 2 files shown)


leplatrem:
This looks good to me! Thanks for this contribution! 

I let @ametaireau have a look and merge it :)

ametaireau:
That looks good to me as well! Thanks a lot :+1:

tsauerwein:
Thanks! :)

prmtl:
Are there any plans on making a new release that will include this fix?

PR #344, opened 2015-11-27
add Resource predicates support

vmaksymiv:
It allows to register different Resources with the same path.
Ex.

```
@resource(name='FlyingUnicorn', path='/unicorns/{lovely_name}', unicorn_aptitude='fly')
class FlyingUnicorn():
    @view
    def get(self):
         # FlyingUnicorn stuff

@resource(name='MagicUnicorn', path='/unicorns/{lovely_name}', unicorn_aptitude='do magic')
class MagicUnicorn():
    @view
    def get(self):
         # MagicUnicorn stuff
```


leplatrem:
Thanks for your contribution!

The feature does not look irrelevant to me. I witl let @almet confirm, but it is sure that it can't land in master without tests :)

almet:
I confirm, this would be useful, but needs testing.

leplatrem:
@vmaksymiv would you have some time to contribute a couple of tests ?

vmaksymiv:
Hello!
Sorry for the late response.
We added tests for "custom predicates" feature and they successfully passed. 
But to make all test checks work we excluded from coverage one line (explained in comment to that line).
We couldn't find parameters set for request to reach this line. Maybe you can find better solution than excluding the line?

leplatrem:
Great! Thanks for your additional efforts on this!

Indeed, I agree with your comment. This particular line is impossible to reach, and didn't feel confident enough to get rid of it during the 2.0 refactor (See https://github.com/Cornices/cornice/blob/master/tests/test_pyramidhook.py#L280-L289).

PR #346, opened 2016-01-18
Remove superfluous call to get_localizer (ref #206)

leplatrem:
See https://github.com/Pylons/pyramid/blob/1.5.8/pyramid/i18n.py#L211-L218

@amarandon any objection :) ?


amarandon:
`request.localizer` is used in schema.py and I'm not sure where it comes from now, but [that line](https://github.com/mozilla-services/cornice/blob/master/cornice/schemas.py#L151) is covered by tests and still passing, so should be OK :)

PR #348, opened 2016-02-03
Incorrect attribute 'validator' replaced with 'validators'

peletiah:
This typo in [Testing example](https://cornice.readthedocs.org/en/latest/testing.html) caused pserve to fail with `pyramid.exceptions.ConfigurationError: Unknown predicate values: {'validator': <function has_payed at 0x7fb14c54f500>}`


almet:
Thanks a lot :)

PR #350, opened 2016-02-28
properly handle content_type callables returning a single internet media type as scalar

### LINKED ISSUE #343: content_type not handled properly if it is a function
I have defined `view` decorator in my `resource`:

    @view(renderer='json', content_type='application/json')

and when I call it without `Content-Type` properly defined, I get a 415 error.

However, if I want to call a function instead of defining a string for `content_type`

    @view(renderer='json', content_type=content_type)

and in some other file i have defined that function:

    def content_type(request):
        # interact with request
        return 'application/json'

it won't produce error 415, but it will return a 200 response.

Why is this happening? For some reason I'm sure it was working few months ago when I defined it in this way...

amotl:
Improve and resolve #343, add appropriate tests.

### DIFF
--- a/cornice/tests/test_validation.py
+++ b/cornice/tests/test_validation.py
@@ -94,6 +94,16 @@ class TestServiceDefinition(LoggingCatcher, TestCase):
         response = app.get('/service3', headers={'Accept': 'text/*'})
         self.assertEqual(response.content_type, "text/json")
 
+        # test that using a callable to define what's accepted works as well
+        # now, the callable returns a scalar instead of a list
+        response = app.put('/service3', headers={'Accept': 'audio/*'},
+                           status=406)
+        error_description = response.json['errors'][0]['description']
+        self.assertIn('text/json', error_description)
+
+        response = app.put('/service3', headers={'Accept': 'text/*'})
+        self.assertEqual(response.content_type, "text/json")
+
         # if we are not asking for a particular content-type,
         # we should get one of the two types that the service supports.
         response = app.get('/service2')
@@ -243,7 +253,14 @@ class TestServiceDefinition(LoggingCatcher, TestCase):
         self.assertTrue('text/xml' in error_description)
         self.assertTrue('application/json' in error_description)
 
-        app.post('/service6', headers={'Content-Type': 'text/xml'})
+    def test_content_type_with_callable_returning_scalar(self):
+        # test that using a callable for content_type works as well
+        # now, the callable returns a scalar instead of a list
+        app = TestApp(main({}))
+        response = app.put('/service6', headers={'Content-Type': 'audio/*'},
+                           status=415)
+        error_description = response.json['errors'][0]['description']
+        self.assertIn('text/xml', error_description)
 
     def test_accept_and_content_type(self):
         # tests that giving both Accept and Content-Type
--- a/cornice/tests/validationapp.py
+++ b/cornice/tests/validationapp.py
@@ -62,11 +62,8 @@ def get2(request):
 service3 = Service(name="service3", path="/service3")
 
 
-def _accept(request):
-    return ('application/json', 'text/json')
-
-
-@service3.get(accept=_accept)
+@service3.get(accept=lambda request: ('application/json', 'text/json'))
+@service3.put(accept=lambda request: 'text/json')
 def get3(request):
     return {"body": "yay!"}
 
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -1,5 +1,7 @@
 .tox
 .hg*
+.venv*
+.idea
```
almet: what is this? (should it be part of **your** .gitignore instead?)

cornice/tests/test_validation.py
```diff
@@ -94,6 +94,16 @@ def test_accept(self):
...
 
+        # test that using a callable to define what's accepted works as well
+        # now, the callable returns a scalar instead of a list
+        response = app.put('/service3', headers={'Accept': 'audio/*'},
+                           status=406)
+        error_description = response.json['errors'][0]['description']
+        self.assertIn('text/json', error_description)
+
+        response = app.put('/service3', headers={'Accept': 'text/*'})
+        self.assertEqual(response.content_type, "text/json")
```
almet: nit: replace the double quotes with simple ones.

cornice/tests/validationapp.py
```diff
@@ -62,11 +62,8 @@ def get2(request):
 service3 = Service(name="service3", path="/service3")
 
 
-def _accept(request):
-    return ('application/json', 'text/json')
-
-
-@service3.get(accept=_accept)
+@service3.get(accept=lambda request: ('application/json', 'text/json'))
+@service3.put(accept=lambda request: 'text/json')
```
almet: might be better to define another service (and name it properly, we shouldn't continue with this bad habit of naming these "serviceX" !)

almet:
Hi, and thanks for doing this! I've put a bunch of comments here and there, let me know when you had the time to address them!

amotl:
Hey Alex, thanks for reviewing this. Will answer the comments inline.

.gitignore
amotl: - ``.idea`` is the project metadata folder of *PyCharm*, which is in pretty common use amongst our folks.
- When working with python virtualenvs, i think it is a convention to use ``.venv27`` or ``.venv34`` folders for designating the type of virtualenv. Given, when using *tox* one usually doesn't need manual-made virtualenvs for running the tests, but we usually even install *tox* into a project-local *virtualenv* to complete decouple from the system python.

Both entries shouldn't do any harm but instead should gain value for others, so they were added on purpose. However, we can remove them again if they make anyone sad.

cornice/tests/test_validation.py
amotl: Will do. The lines have been copied from the other test cases which are overall somehow not in the best shape (see next comment of yours and my minor improvements in #351). Will be happy to improve some aspects of that slightly as we go when having the opportunity if you would appreciate that.

cornice/tests/validationapp.py
amotl: - Having the services named ``serviceX`` offended my eyes too when having the first look the other day, so i totally second that.
- However, the ``service3`` is perfectly for testing this case of having a callable as argument to ``accept`` and can well carry both flavors (returning list vs. returning scalar) when renamed properly.

As said, i'm happy to add another PR after merging #350 and #351 to improve the shape of the test cases, especially regarding minor refactoring and proper renaming of the ``serviceX`` jungle.

amotl:
Hi Alex, i tried to clarify my intentions and give an outlook. Let me know if there's still any blocking issue for you in this, otherwise i would appreciate doing a different PR for more cosmetic and cleanup work centered around the things mentioned. Best, Andreas.

.gitignore
almet: I believe these two lines should not be included because they're specific to one setup. It would make this file harder to maintain over time, for little benefit :-)

cornice/tests/test_validation.py
almet: thanks :)

almet:
I think we're good once the .gitignore lines have been removed!

amotl:
Thanks! The .gitignore lines have been removed by a squashed commit.

almet:
Thanks!

amotl:
Thanks for merging! Would you also have a look at #351? This complements this change as it improves and clarifies the documentation regarding this topic at the places @bomb-on and @leplatrem were looking at in the context of #343.

PR #351, opened 2016-02-28
Improve documentation regarding content type negotiation and media type validation

amotl:
Improve documentation of content type negotiation and media type validation. See also #91, #343, #350.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 .tox
 .hg*
+.venv*
+.idea
 *.pyc
 *.mako.py
 lib
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,5 @@
 .tox
 .hg*
-.venv*
-.idea
 *.pyc
 *.mako.py
 lib
--- a/cornice/tests/test_validation.py
+++ b/cornice/tests/test_validation.py
@@ -59,9 +59,9 @@ class TestServiceDefinition(LoggingCatcher, TestCase):
         error_description = response.json['errors'][0]['description']
         self.assertEquals('header', error_location)
         self.assertEquals('Accept', error_name)
-        self.assertTrue('application/json' in error_description)
-        self.assertTrue('text/json' in error_description)
-        self.assertTrue('text/plain' in error_description)
+        self.assertIn('application/json', error_description)
+        self.assertIn('text/json', error_description)
+        self.assertIn('text/plain', error_description)
 
         # requesting a supported type should give an appropriate response type
         response = app.get('/service2', headers={'Accept': 'application/*'})
@@ -89,13 +89,13 @@ class TestServiceDefinition(LoggingCatcher, TestCase):
         response = app.get('/service3', headers={'Accept': 'audio/*'},
                            status=406)
         error_description = response.json['errors'][0]['description']
-        self.assertTrue('text/json' in error_description)
+        self.assertIn('text/json', error_description)
 
         response = app.get('/service3', headers={'Accept': 'text/*'})
         self.assertEqual(response.content_type, "text/json")
 
-        # test that using a callable to define what's accepted works as well
-        # now, the callable returns a scalar instead of a list
+        # Test that using a callable to define what's accepted works as well.
+        # Now, the callable returns a scalar instead of a list.
         response = app.put('/service3', headers={'Accept': 'audio/*'},
                            status=406)
         error_description = response.json['errors'][0]['description']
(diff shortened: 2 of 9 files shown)


.gitignore
```diff
@@ -1,5 +1,7 @@
 .tox
 .hg*
+.venv*
+.idea
```
almet: heh :-)

cornice/tests/test_validation.py
```diff
@@ -89,16 +89,26 @@ def test_accept(self):
         response = app.get('/service3', headers={'Accept': 'audio/*'},
                            status=406)
         error_description = response.json['errors'][0]['description']
-        self.assertTrue('text/json' in error_description)
+        self.assertIn('text/json', error_description)
```
almet: Thanks for doing this: we'll have better output in case of errors!

cornice/tests/test_validation.py
```diff
@@ -89,16 +89,26 @@ def test_accept(self):
...
 
         response = app.get('/service3', headers={'Accept': 'text/*'})
         self.assertEqual(response.content_type, "text/json")
 
+        # test that using a callable to define what's accepted works as well
```
almet: nit: Sentences should start with a cap and end with a dot.

cornice/tests/test_validation.py
```diff
@@ -240,10 +250,17 @@ def test_content_type_with_callable(self):
...
+        self.assertIn('application/json', error_description)
 
-        app.post('/service6', headers={'Content-Type': 'text/xml'})
+    def test_content_type_with_callable_returning_scalar(self):
+        # test that using a callable for content_type works as well
+        # now, the callable returns a scalar instead of a list
+        app = TestApp(main({}))
+        response = app.put('/service6', headers={'Content-Type': 'audio/*'},
+                           status=415)
+        error_description = response.json['errors'][0]['description']
```
almet: after seing this twice, I wonder if we should make it a helper function that we could put in cornice/tests/support.py instead. What do you think ? (this can be done in a later pull request if you want)

cornice/tests/validationapp.py
```diff
@@ -50,6 +50,7 @@ def post1(request):
     return {"body": request.body}
 
 
+# service for testing the "accept" parameter (scalar and list)
```
almet: nit: Sentences should start with a cap and end with a dot.

cornice/util.py
```diff
@@ -104,22 +104,42 @@ def json_error(errors):
 
 def match_accept_header(func, context, request):
     """
-    Return True if the request matches the values returned by the given :param:
-    func callable.
+    Return True if the request Accept headers match
+    the list returned by callable in :param:func.
```
almet: `by the callable specified in :param:`

cornice/util.py
```diff
@@ -104,22 +104,42 @@ def json_error(errors):
...
+
+    Also attach the total list of possible accepted
+    egress media types to the request.
```
almet: I don't understant the use of "egress" in this sentence :x

almet: Got it after reading a bit more the docs.

docs/source/testing.rst
```diff
@@ -1,6 +1,25 @@
...
+
+    tox
+
+To run a single test inside a specific Python environment, do e.g.::
+
+    tox -e py27 cornice/tests/test_validation.py:TestServiceDefinition.test_content_type_missing
+
+or::
+
+    tox -e py27 cornice.tests.test_validation:TestServiceDefinition.test_content_type_missing
```
almet: Great!

almet:
This is really great, thanks for taking to time to do this, especially adding documentation, that will (I'm sure) be useful to a few others.

I've added a few nitpicks here and there. Let me kow when they've been addressed, and I think we should be good to go! (this branch needs to be rebased on master as well).

amotl:
Hi Alex, thanks again for reviewing this. I will have a look into the topics and amend things accordingly. Sorry for the glitch re. .gitignore: Missed on rebasing this branch on top of the one of yesterday #350 before calling out for review.

cornice/tests/test_validation.py
amotl: I'm thinking the same. Would it be fine to define the services required for testing all the nitty-gritty details completely at runtime? When considering this, i can imagine targeting these topics as eventual goals of the test cleanup task:

- The */serviceX* jungle in ``validationapp.py`` would be completely cleaned up, the services will be defined at runtime from the particular test case in ``test_validation.py`` by using a helper function from ``cornice/tests/support.py``, as you suggested. When we're already at spawning the appropriate service at test execution time, let's also revamp the uri namespace to something like */test/service/accept* (former uris */service2* and */service3*) or */test/service/content-type* (former */service5*), to name just two examples to improve on "naming things".
- Just a **single** decorator-based service definition would remain inside the revamped ``validationapp.py`` for testing the decorator machinery against all available http verbs on a single service uri */test/service/decorated*.

Does this sound good to you? As we didn't define services at runtime before and were just using the decorator-style definition in our projects, some hints about doing this would definitively help. Currently, i can't get my head around **not** using Venusian for scanning the service decorator definitions. Thanks in advance!

amotl:
Polished as suggested and rebased on current master.

leplatrem:
This is great!

Thank you very much for your efforts on this!

amotl:
Dear Mathieu, thanks for merging and big up to the Mozilla Services French Team for its awesome work on bringing REST at its finest to the Pyramid ecosystem. Looking forward to playing with Cliquet and Kinto.

almet:
Thanks @amotl for the kind words. Thank you for your help on these pull requests, that's really appreciated.

PR #352, opened 2016-03-01
Clarify docs for deferred colander validator

antoineleclair:
Without the decorator, the second argument is not `kw` but the actual value of the node.


almet:
Thanks for this :)

PR #354, opened 2016-03-14
Do not raise a 415 error when no content-type and no body

leplatrem:
Return a 415 error response, if and only if there is some content.

@almet r?

### DIFF
--- a/cornice/pyramidhook.py
+++ b/cornice/pyramidhook.py
@@ -219,8 +219,8 @@ def register_service_views(config, service):
         predicate_definitions = _pop_complex_predicates(args)
 
         if predicate_definitions:
-            extra_empty_ctype = [({'kind': 'content_type', 'value': ''},)]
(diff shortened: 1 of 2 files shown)


Natim:
Refs https://github.com/mozilla-services/cornice/issues/353

Natim:
I did some manual checking with Kinto and the behavior is the one we want :+1:

cornice/pyramidhook.py
```diff
@@ -219,7 +219,8 @@ def register_service_views(config, service):
         predicate_definitions = _pop_complex_predicates(args)
 
         if predicate_definitions:
-            for predicate_list in predicate_definitions:
+            extra_empty_ctype = [({'kind': 'content_type', 'value': ''},)]
```
almet: Ctype [refers to something different](https://docs.python.org/2/library/ctypes.html) in the python context.

almet:
That's indeed a better behaviour. We should change the naming, but apart from that it looks good to land!

PR #359, opened 2016-04-16
Convert None to colander.null before calling deserialize

### LINKED ISSUE #349: None values in nested schemas
When using nested schemas, there is a problem during the validation of data that contains `None` values. For example for the following schema, the validation fails with the error message `colander.Invalid: {'b.a': '"None" is not a mapping type: Does not implement dict-like functionality.'}`.

    class SchemaA(MappingSchema):
        val = SchemaNode(String())
    class SchemaB(MappingSchema):
        a = SchemaA()
    class SchemaC(MappingSchema):
        b = SchemaB()

    schema = CorniceSchema.from_colander(SchemaC)
    dummy_request = get_mock_request( '{"b": {"a": null}}')
    validate_colander_schema(schema, dummy_request)

Similar errors arise when using sequences or tuples in nested mappings (`'"None" is not iterable'`). I added 3 failing tests to the test-suite to demonstrate the issue: 
https://github.com/tsauerwein/cornice/commit/5bd09404d10fad5b7065c0a98a7bc250ff47e360

The problem is that [deserialize](https://github.com/mozilla-services/cornice/blob/0dda4379e51ffb902b4ded26023e4ae2fa50bae0/cornice/schemas.py#L148) is called with data that still contains `None` values. Because Colander expects `null`, it will error during the validation.

I would propose to recursively replace all `None`s with `null` before calling `deserialize`. I can prepare a PR, but I first wanted to check if there is no better solution.

tsauerwein:
This PR is an attempt to fix https://github.com/mozilla-services/cornice/issues/349 by replacing all occurrences of `None` with `colander.null` before the validation.

A more detailed description of the problem is in #349.

Closes #349

### DIFF
--- a/cornice/validators/_colander.py
+++ b/cornice/validators/_colander.py
@@ -1,7 +1,9 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
-
+import colander
+import collections
+import webob.multidict
 import inspect
 import warnings
 
@@ -68,7 +70,8 @@ def validator(request, schema=None, deserializer=None, **kwargs):
         return
 
     schema = _ensure_instantiated(schema)
-    cstruct = deserializer(request)
+    cstruct = _none_to_null(deserializer(request))
+
     try:
         deserialized = schema.deserialize(cstruct)
     except colander.Invalid as e:
@@ -90,3 +93,28 @@ def _ensure_instantiated(schema):
             stacklevel=2)
         schema = schema()
     return schema
+
+
+def _none_to_null(data):
+    """Replace all `None` values with `colander.null` in the given
+    structure because `deserialize` expects `colander.null`.
+    """
+    if data is None:
+        return colander.null
+    elif not data:
+        return data
+    elif isinstance(data, webob.multidict.MultiDict):
+        new_data = webob.multidict.MultiDict()
+        for key in data:
+            values = data.getall(key)
+            for value in values:
+                new_data[key] = _none_to_null(value)
+        data = new_data
+    elif isinstance(data, collections.Mapping):
+        for key in data:
+            data[key] = _none_to_null(data[key])
+    elif isinstance(data, list):
+        data = [_none_to_null(value) for value in data]
+    elif isinstance(data, set):
+        data = (_none_to_null(value) for value in data)
+    return data
--- a/tests/support.py
+++ b/tests/support.py
@@ -1,6 +1,7 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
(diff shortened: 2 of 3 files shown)


leplatrem:
I don't really understand why colander is not in charge of this. Is it because of the hacky usage of colander in Cornice?

Otherwise it does look bad to me. @almet @Natim any feedback?

leplatrem:
@tsauerwein do you know why we have to do this? Why colander would not be in charge of this?

tsauerwein:
> Why colander would not be in charge of this?

That is a good question, and I am not totally sure about this myself.

In their tests they are explicitly testing that `None` can not be processed:
https://github.com/Pylons/colander/blob/master/colander/tests/test_colander.py#L935
or https://github.com/Pylons/colander/blob/master/colander/tests/test_colander.py#L690

The documentation of [deserialize](http://colander.readthedocs.io/en/latest/api.html#colander.SchemaNode.deserialize) says: "Deserialize the cstruct into an appstruct ...". And a [cstruct](http://colander.readthedocs.io/en/latest/glossary.html#term-cstruct) is a "data structure generated by the colander.SchemaNode.serialize()". And I assume that a `cstruct` is not supposed to contain `None`?

leplatrem:
The behaviour with `None` does not seem weird:

```py
>>> import colander
>>> class Payload(colander.MappingSchema):
...     name = colander.SchemaNode(colander.String())
... 
>>> class Body(colander.MappingSchema):
...     payload = Payload()
... 
>>> schema = Body()
>>> schema.deserialize({"payload": {"name": None}})
  File "<stdin>", line 1, in <module>
  File "/home/mathieu/Code/Mozilla/kinto/.venv/local/lib/python2.7/site-packages/colander/__init__.py", line 2058, in deserialize
    appstruct = self.typ.deserialize(self, cstruct)
  File "/home/mathieu/Code/Mozilla/kinto/.venv/local/lib/python2.7/site-packages/colander/__init__.py", line 719, in deserialize
    return self._impl(node, cstruct, callback)
  File "/home/mathieu/Code/Mozilla/kinto/.venv/local/lib/python2.7/site-packages/colander/__init__.py", line 699, in _impl
    raise error
colander.Invalid: {'payload.name': u'Required'}
>>>

tsauerwein:
> The behaviour with None does not seem weird:

In this case, yes. But it gets weird once you have sequences or nested schemas. I added some examples in the tests: https://github.com/Cornices/cornice/pull/359/files#diff-bdda56b22b0eb46ed999750ec24c5785R121

leplatrem:
Hum, I see...

Well to be honest I'm currently not sure if Cornice should be in charge of handling the weird cases of Colander :/

I think it would be good to start a discussion with Colander devs first, based a very simple example that reproduces the problem. And then do such preprocessing here only if we're obliged to :) 

(The PR is good, it's just that we're trying to keep the source code small and decorrelated as much as possible from colander specificities)

tsauerwein:
That makes sense, yes. I will try to get an answer from the Colander folks.

leplatrem:
I can see that the issue upstream on Colander is still open.
But I'm closing this PR, since it's 4 years old.

PR #361, opened 2016-06-07
Erase the SERVICES variable after having used it, not before.

maparent:
If the cornice Services are loaded early in the sphinx conf.py, they will not be in MODULES, and the current code will just import them which will not re-register them in SERVICES, so we get an empty services.


Natim:
Moved here: https://github.com/Cornices/cornice.ext.sphinx/pull/3/files

PR #362, opened 2016-06-09
fix travis python version

jptomo:
It seems to not valid python version.

best regards


jptomo:
It's my misunderstanding. sorry.

PR #365, opened 2016-08-06
Port messaging example to python3. Requires requests networking frame…

mazz:
…work.


leplatrem:
Thanks for this contribution! 

I think you can modify the existing files instead of adding new ones. Python3 is not an exotic thing anymore ;)

sidha:
No problem! Glad to help out. I was more concerned with the 'requests' dependency. Are we good with that?

mazz:
I updated  examples/messaging/messaging dir with the python3 examples. Build is failing though(probably because python 2.x is compiling it)

sidha:
I updated examples/messaging/messaging dir with the python3 examples. Build is failing though(probably because python 2.x is compiling it)

examples/messaging/messaging/views.py
```diff
@@ -1,26 +1,24 @@
 """ Cornice services.
 """
+from cornice import Service
```
leplatrem: superfluous I think

examples/messaging/messaging/client.py
```diff
@@ -105,3 +111,5 @@ def shell():
 
 if __name__ == '__main__':
     shell()
+
+
```
leplatrem: Why empty lines?

examples/messaging/messaging/client.py
```diff
@@ -1,36 +1,39 @@
 import threading
-import urllib2
+import requests
```
leplatrem: This must probably be added to the setup dependencies

examples/messaging/messaging/client.py
```diff
@@ -1,36 +1,39 @@
...
+from requests import Request, Session
+from requests.auth import HTTPBasicAuth
+from requests.packages.urllib3.exceptions import InsecureRequestWarning
+from requests import RequestException
+
+requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
```
leplatrem: Not sure about this... let's keep this client code as simple as possible

examples/messaging/messaging/client.py
```diff
@@ -1,36 +1,39 @@
...
-    message = {'text': message}
-    req.add_data(json.dumps(message))
-    urllib2.urlopen(req)
-
+    data = {'text': message}
+    s = Session()
+    req = Request('POST', _SERVER, json=data)
+    prepped = s.prepare_request(req)
+    prepped.headers['X-Messaging-Token'] = token
+    resp = s.send(prepped)
```
leplatrem: Why not just `requests.post(_SERVER, data=data, headers={'X-Messaging-Token': token})` ?

examples/messaging/messaging/client.py
```diff
@@ -1,36 +1,39 @@
...
     url = _SERVER + '/users'
-    req = urllib2.Request(url)
-    req.add_data(name)
     try:
-        res = urllib2.urlopen(req)
-    except urllib2.HTTPError:
+        s = Session()
+        req = Request('POST', url, data=name)
+        prepped = s.prepare_request(req)    
+        resp = s.send(prepped)
```
leplatrem: ditto: the `requests` lib is a lot simpler than that code :)

examples/messaging/messaging/client.py
```diff
@@ -1,36 +1,39 @@
...
+    except RequestException:
         return False
 
-    if res.getcode() != 200:
+    if resp.status_code != 200:
         return False
 
-    return json.loads(res.read())['token']
-
+    return json.loads(resp.text)['token']
```
leplatrem: Use `resp.json()` instead

examples/messaging/messaging/client.py
```diff
@@ -43,12 +46,15 @@ def __init__(self, server, token, scr):
...
+        req = Request('GET', self.server)
+        prepped = s.prepare_request(req)
+        prepped.headers['X-Messaging-Token'] = self.token
+        resp = s.send(prepped)
 
         while self.updating:
-            res = urllib2.urlopen(req)
-            result = json.loads(res.read())
+            r = s.get(self.server)
+            result = json.loads(resp.text)
```
leplatrem: Use `resp.json()` instead

examples/messaging/messaging/client.py
sidha: Without these imports, the output becomes full with many exceptions

Natim: Actually if you've got warning it is because your installation of Python is not secure.
You should install `pip install PyOpenSSL cryptography idna  certifi`

Or simply `pip install urllib3[secure]`

See https://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings

PR #369, opened 2016-08-17
Get rid of buildout stuff

### LINKED ISSUE #177: Supplied bootstrap.py is not Python 3 Compatible
When I try to run bootstrap.py I get a syntax error. It seems that the bootstrap provided does not work with Python 3 (that is 3.3).

leplatrem:
Buildout does not seem to bring any concrete value here, so remove it.

Fixes #177


Natim:
Don't hesitate to update the CHANGELOG as well. Support for it was added in 2012.

PR #372, opened 2016-08-23
Remove request from errors list (fixes #328)

### LINKED ISSUE #328: circular reference created for every request.
from pyramidhook.py:
```python
  def wrap_request(event):
      """Adds a "validated" dict, a custom "errors" object and an "info" dict to
      the request object if they don't already exists
      """
      request = event.request
      request.add_response_callback(apply_filters)

      if not hasattr(request, 'validated'):
          setattr(request, 'validated', {})

      if not hasattr(request, 'errors'):
          setattr(request, 'errors', Errors(request))

      if not hasattr(request, 'info'):
          setattr(request, 'info', {})
```
...

from errors.py:
```python
  class Errors(list):
      """Holds Request errors
      """
      def __init__(self, request=None, status=400):
          self.request = request
          self.status = status
          super(Errors, self).__init__()
```
...

Thus request->errors->request

leplatrem:
The errors list is accessed via ``request.errors``, it makes thus no sense to store a circular reference.

And «Look mum, no tests fail!» (*ie. no spec was changed*)

@Natim r?

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -9,6 +9,7 @@ CHANGELOG
 Breaking changes
 
 - Get rid of Buildout files (#369)
+- Errors list ``request.errors`` has no ``request`` anymore (fixes #328)
 
 
 1.2.1 (2016-03-15)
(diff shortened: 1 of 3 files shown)


Natim:
r+

Natim:
It looks like we need it in Kinto at least: https://github.com/Kinto/kinto/issues/800

PR #374, opened 2016-08-23
Only accept known locations in request.errors.add() (fixes #99)

### LINKED ISSUE #99: Check that the errors contain the right parameters
For instance, we should cry if we put "headers" instead of "header" when doing `request.errors.add()`.

leplatrem:
If this conflicts with your PR @tarekziade, do not take it into account, I'll rebase it later.
r?

### DIFF
--- a/cornice/errors.py
+++ b/cornice/errors.py
@@ -16,6 +16,11 @@ class Errors(list):
 
     def add(self, location, name=None, description=None, **kw):
         """Registers a new error."""
+        allowed = ('body', 'querystring', 'url', 'header', 'path',
+                   'cookies', 'method')
+        if location not in allowed:
+            raise ValueError('%r not in %s' % (location, allowed))
+
         self.append(dict(
             location=location,
             name=name,
--- a/cornice/errors.py
+++ b/cornice/errors.py
@@ -18,7 +18,7 @@ class Errors(list):
         """Registers a new error."""
         allowed = ('body', 'querystring', 'url', 'header', 'path',
                    'cookies', 'method')
-        if location not in allowed:
+        if location != '' and location not in allowed:
             raise ValueError('%r not in %s' % (location, allowed))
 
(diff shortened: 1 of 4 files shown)


leplatrem:
From what I see in #376, it should probably also support `cookie`, `path`, `method`

leplatrem:
@Natim r?

cornice/tests/test_errors.py
```diff
@@ -0,0 +1,12 @@
...
+from cornice.errors import Errors
+
+
+class TestErrorsHelper(TestCase):
+    def setUp(self):
+        self.errors = Errors()
+
+    def test_raises_an_exception_when_location_is_unsupported(self):
+        with self.assertRaises(ValueError):
+            self.errors.add('something')
```
Natim: Can we add tests for each accepted parameters?

cornice/errors.py
```diff
@@ -18,7 +18,7 @@ def add(self, location, name=None, description=None, **kw):
         """Registers a new error."""
         allowed = ('body', 'querystring', 'url', 'header', 'path',
                    'cookies', 'method')
-        if location not in allowed:
+        if location != '' and location not in allowed:
```
Natim: Do you really want to accept empty location?

cornice/errors.py
```diff
@@ -16,6 +16,11 @@ def __init__(self, status=400):
 
     def add(self, location, name=None, description=None, **kw):
         """Registers a new error."""
+        allowed = ('body', 'querystring', 'url', 'header', 'path',
+                   'cookies', 'method')
+        if location != '' and location not in allowed:
```
Natim: Do you really want to accept empty location?

leplatrem: That's what colander gives when you raise a validation error that is not related to a specific node (eg. `provide param either via header or querystring`)

PR #375, opened 2016-08-23
Scaffold now works when project name contains dashes (fixes #238)

### LINKED ISSUE #238: Bug when package name has a dash
Pyramid's .ini scaffolds have ```use = egg:{{project}}``` in them, but cornice/scaffolds/cornice/+package+.ini_tmpl has ```use = egg:{{package}}```. This creates a .ini file with an incorrect egg reference, which causes a ```pkg_resources.DistributionNotFound``` exception when pserve is run. I've tested that replacing {{package}} with {{project}} does fix this issue, and I can only assume that nothing else will break with this change.

leplatrem:
@Natim r?

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -10,6 +10,10 @@ Breaking changes
 
 - Get rid of Buildout files (#369)
 
+Bug fixes
+
+- Scaffold now works when project name contains dashes (fixes #238)
+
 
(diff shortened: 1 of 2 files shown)


Natim:
Shall we add a test for that?

Natim:
If you think we don't r+

Issue #379, opened 2016-09-08
Remove cornice.ext & move to a separate org


tarekziade:
Looking at  https://github.com/mozilla-services/cornice/pull/340 and https://github.com/mozilla-services/cornice/pull/373

I am wondering what's the real value of keeping swagger, spore, \<any generator\> integration inside cornice.  It seems simple enough to have them packaged at PyPI under cornice-xxx and just use them.

Also, maybe we could have cornice becoming a github org and host cornice-related packages alongside the cornice package there, so people can work in their integration there and be part of the Cornice project.

What do you think?

Natim:
I tend to agree with you on this. It works pretty well like that for kinto. :+1:

leplatrem:
I agree with that!

An alternative to creating a dedicated org -- in case it looks overkill -- would be to have a section in the docs with links to cornice related projects.

leplatrem:
Done

PR #383, opened 2016-09-28
Removed cornice.ext

tarekziade:


leplatrem:
Maybe add some context and explanation in the PR description (ref to issue) :)

r+ otherwise

leplatrem:
Well, it was issue #379 and they are going to be moved to separate python packages

PR #389, opened 2016-10-18
Fix docs from tox

leplatrem:
*Every PR fails because of this*.

@Natim r?


glasserc:
Looks OK to me. Did this ever work?

PR #398, opened 2016-10-20
Move tests out of package

leplatrem:
@Natim r?


tests/validationapp.py
```diff
@@ -1,12 +1,14 @@
...
+import json
+
 from pyramid.config import Configurator
 from pyramid.httpexceptions import HTTPBadRequest
 
 from cornice import Service
-from cornice.tests.support import CatchErrors
-import json
+
+from .support import CatchErrors
```
Natim: Are you sure that we don't need that to test extensions for instance?

leplatrem: Yes, and if it would be the case, it should live in some `cornice.testing` module

Issue #411, opened 2016-10-26
Semantics of validation from ``path`` has changed


dairiki:
In Cornice < 2, when a schema item had ``location=path``, it was extracted/validated from ``request.matchdict``.  This allowed for code like

```python
class MySchema(colander.Schema):
    item_id = colander.SchemaNode(
        colander.Integer(),
        location='path')

@resource(
    path='/items/{item_id}.json',
    schema=MySchema)
class ItemResource(object):
    def __init__(self, request):
        self.request = request

    def get(self):
        item_id = request.validated['item_id']
        return {'id': item_id}
```

In Cornice >= 2, I would naively translate the above to

```python
class PathSchema(colander.Schema):
    item_id = colander.SchemaNode(colander.Integer())

class MySchema(colander.Schema):
    path = PathSchema()

@resource(
    path='/items/{item_id}.json',
    schema=MySchema,
    validators=(colander_validator,))
class ItemResource(object):
    def __init__(self, request):
        self.request = request

    def get(self):
        item_id = request.validated['path']['item_id']
        return {'id': item_id}
```

This does not, however, seem to work.  Currently, ``extract_cstruct`` puts ``request.path`` (a string) into ``cstruct['path']`` (and does not appear to make ``request.matchdict`` available at all.)

I would suggest that ``extract_cstruct`` should be modified to make ``request.matchdict`` available in the cstruct; either as ``cstruct['path']`` (replacing ``request.path``) or as ``cstruct['matchdict']``.

Am I missing something?

leplatrem:
Your feedback seems totally valid!

In Cornice 1.2, we used `request.matchdict`
https://github.com/Cornices/cornice/blob/1.2.1/cornice/util.py#L187

I'm open to both approaches: `request['path']` as before, or in a new key `'matchdict'`. The first one would just be a simple bug fix, the other one is a new feature with the required details in the upgrading docs.

Would you be willing to open a pull-request with the small fix please?

dairiki:
Yes, I can come up with a pull request.

(Upon further consideration, I think my first approach is the right one.  Otherwise the semantics of 'path' change for error reporting as well.)

Natim:
However I thought the change for the error reporting part was done on purpose.

PR #413, opened 2016-10-26
Restore old semantics of location='path'

dairiki:
Change ``extract_cstruct`` so as to put ``request.matchdict`` (rather than ``request.path``) into ``cstruct['path']``.

This is to maintain semantic consistency with Cornice 1.2, where ``location='path'`` meant to extract parameters from ``matchdict``.

See #411 for further discussion.


leplatrem:
Thanks a lot!

PR #416, opened 2016-10-27
Make colander_validator a no-op if no schema specified

### LINKED ISSUE #414: Clarify colander_validator behavior when no schema is specified.
The docstring for ``colander_validator`` states “If no schema is defined, this validator does nothing”, however, currently, an exception is raised if ``schema`` is ``None``.

https://github.com/Cornices/cornice/blob/master/cornice/validators/_colander.py#L47
https://github.com/Cornices/cornice/blob/master/cornice/validators/_colander.py#L63

Which is the desired behavior?

dairiki:
Fixes #414

### DIFF
--- a/cornice/validators/_colander.py
+++ b/cornice/validators/_colander.py
@@ -62,28 +62,19 @@ def validator(request, schema=None, deserializer=None, **kwargs):
     if deserializer is None:
         deserializer = extract_cstruct
 
-    if schema is None:
-        raise TypeError('This validator cannot work without a schema')
-
-    schema = _ensure_instantiated(schema)
-    cstruct = deserializer(request)
-    try:
-        deserialized = schema.deserialize(cstruct)
-        request.validated.update(deserialized)
-    except colander.Invalid as e:
-        translate = request.localizer.translate
-        error_dict = e.asdict(translate=translate)
-        for name, msg in error_dict.items():
-            prefixed = name.split('.', 1)
-
-            if len(prefixed) == 1:
-                location = name
-                field = ''
-            else:
-                location = prefixed[0]
-                field = prefixed[1]
-
-            request.errors.add(location, field, error_dict[name])
+    if schema is not None:
+        schema = _ensure_instantiated(schema)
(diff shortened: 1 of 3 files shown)


cornice/validators/_colander.py
```diff
@@ -62,28 +62,19 @@ def validator(request, schema=None, deserializer=None, **kwargs):
...
-
-            if len(prefixed) == 1:
-                location = name
-                field = ''
-            else:
-                location = prefixed[0]
-                field = prefixed[1]
-
-            request.errors.add(location, field, error_dict[name])
+    if schema is not None:
```
leplatrem: nitpick: to prevent indenting the whole block, test if None and exit function?

cornice/validators/_colander.py
```diff
@@ -62,28 +62,19 @@ def validator(request, schema=None, deserializer=None, **kwargs):
...
+        schema = _ensure_instantiated(schema)
+        cstruct = deserializer(request)
+        try:
+            deserialized = schema.deserialize(cstruct)
+        except colander.Invalid as e:
+            translate = request.localizer.translate
+            error_dict = e.asdict(translate=translate)
+            for name, msg in error_dict.items():
+                location, _, field = name.partition('.')
+                request.errors.add(location, field, msg)
```
leplatrem: 👍

cornice/validators/_colander.py
dairiki: I come from the days of *Structured Programming* and tend to adhere to the “Single entry, single exit” (SESE) paradigm — it's just the way my brain works.  In this case, there's so little code that it doesn't really matter either way.  I'll change it...

dairiki:
It appears the the reduction in total lines of code in the most recent commit (d701120) has caused a reduction in the test coverage percentage, thus causing coveralls to fail — thus blocking automatic merging.  (Note that the coverage in the edited files is 100%.)

Natim:
@dairiki Don't worry for the coverage :)

PR #417, opened 2016-10-28
Fix coverage

leplatrem:
I'm still not super happy about the way the tests are written, but this is an improvement.

Do you think we should now require 100% for tests to pass?

@Natim r?


Natim:
I tend to think that a 100% coverage for Python project is a minimum.
However I think it isn't 100% yet on Py34.

cornice/pyramidhook.py
```diff
@@ -313,15 +313,10 @@ def _mungle_view_args(args, predicate_list):
...
-            func = callable_map.get(kind)
-            if func:
-                predicate_checker = functools.partial(func, value)
-                predicates.append(predicate_checker)
-                args['custom_predicates'] = predicates
-            else:
-                raise ValueError(
-                    'No function defined for ' +
-                    'handling callables for field "{0}"'.format(kind))
+            func = callable_map[kind]
```
Natim: Are you turning a ValueError into a KeyError there? I think the previous error message was more meaningful about how to solve it that the one we will have with the KeyError.

cornice/errors.py
```diff
@@ -1,10 +1,7 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
-try:
-    import simplejson as json
-except ImportError:
-    import json
+import simplejson as json
```
Natim: Are we still supporting Python2.6? Shouldn't we remove simplejson and use json instead?

Natim: > simplejson is the externally maintained development version of the
json library included with Python 2.6 and Python 3.0, but maintains
backwards compatibility with Python 2.5.

cornice/pyramidhook.py
leplatrem: The previous error cannot happen from what I could understand from this pink floydish code... so this is enough IMO

.coveragerc
```diff
@@ -0,0 +1,2 @@
+[report]
+show_missing = True
```
Natim: 👍

cornice/errors.py
Natim: #418

cornice/pyramidhook.py
Natim: What do you mean it cannot happen?

`callable_map.get(kind)` will return None if the kind is not found, so `if func` will be false and go to the else close. Am I missing something?

leplatrem: I spent a good amount of time trying to raise that error.

Where does `kind` come from? Look well, it can only be `accept` or `content_type`....

leplatrem: `_pop_complex_predicates()` returns a product of `accept` and `content_type`, and `predicate_list` is the result from it plus `empty_contenttype`...

Natim: Ok I trust you on this :)

PR #422, opened 2016-12-10
Fix parsing of 'Access-Control-Request-Headers' header

spoof:
Cornice parses this header by using the non flexible way as HTTP specification
required. Specifications says Access-Control-Request-Headers header
can have any number of LWS (Linear White Spaces) between commas.
But Cornice strictly waits for 'comma with space' delimiter.
It leads to an error for header values like
 `Access-Control-Request-Headers: header1,header2,header3` when 
`cors_expose_all_headers=True`, even if any of these headers (or all of them) 
persist in `service.cors_headers` attribute.

Is it necessary to write tests for this situation? (I see no such tests for other headers)

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -6,7 +6,11 @@ CHANGELOG
 2.3.0 (unreleased)
 ==================
 
-- Nothing changed yet.
+**Bug fixes**
+
+- Fix ``cornice.cors.get_cors_preflight_view`` to make it parse
+  `Access-Control-Request-Headers` header correctly event if its value
+  contains zero number of white spaces between commas (#422)
 
 
 2.2.0 (2016-11-25)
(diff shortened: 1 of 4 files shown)


leplatrem:
Thanks for fixing this!

> Is it necessary to write tests for this situation? (I see no such tests for other headers)

If you can, it would be excellent! Especially to make the spec explicit with a simple test (e.g. reproduce with the example you gave in the description). Some parts of Cornice are not tested with a lot of precision indeed, but let's improve!

Also, could you add a line in the CHANGELOG and add yourself to the contributors please?

spoof:
I've updated my PR: added line to CHANGES.txt and CONTRIBUTORS.txt
I also added testing case to
the `tests/test_cors.py:TestCORS.test_preflight_request_headers_isnt_too_permissive` test.
By doing this I wanted to point that if your if your service is configured with `cors_expose_all_headers=False`, as `eggs` service did, and if you make request with valid headers you get OK and error if they are not valid.
Now we have two cases in test (valid and invalid headers) for service with hiding headers (`cors_expose_all_headers=False` flag) and it's good place to test 'Access-Control-Request-Headers' header value parsing and also check either this flag affects parsing or not.
Is it okay?

PR #423, opened 2016-12-14
Add a request.current_service attribute (fixes #105)

### LINKED ISSUE #105: Add request.current_service attribute

leplatrem:


tests/test_util.py
```diff
@@ -16,3 +16,20 @@ def test_extract_form_urlencoded_data_is_deprecated(self):
...
             util.extract_form_urlencoded_data(mock.MagicMock())
             self.assertTrue(mocked.warn.called)
+
+
+class CurrentServiceTest(unittest.TestCase):
+
+    def test_current_service_returns_the_service_for_existing_patterns(self):
+        request = mock.MagicMock()
+        request.matched_route.pattern = '/buckets'
+        request.registry.cornice_services = {'/buckets': mock.sentinel.service}
```
Natim: Don't you want to have a more functional test there?

PR #424, opened 2016-12-14
Add custom JSON content-type support

gabisurita:
This adds support to a [series of JSON Content-Type variations](http://www.iana.org/assignments/media-types/media-types.xhtml#application) which can be validated as common JSON requests.

Original issue: https://github.com/Kinto/kinto/issues/979
Related to: https://github.com/Kinto/kinto/pull/981

r? @leplatrem

### DIFF
--- a/cornice/validators/__init__.py
+++ b/cornice/validators/__init__.py
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 import re
+
 from webob.multidict import MultiDict
 from cornice.validators._colander import (
     validator as colander_validator,
@@ -31,11 +32,11 @@ def extract_cstruct(request):
     :returns: A mapping containing most request attributes.
     :rtype: dict
     """
+    is_json = re.match('^application/(.*?)json$', str(request.content_type))
+
     if request.content_type == 'application/x-www-form-urlencoded':
         body = request.POST.mixed()
-
-    elif (request.content_type and
-          not re.match('^application/(.*?)json$', str(request.content_type))):
+    elif request.content_type and not is_json:
         body = request.body
     else:
         if request.body:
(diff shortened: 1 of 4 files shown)


cornice/validators/__init__.py
```diff
@@ -33,7 +34,8 @@ def extract_cstruct(request):
     if request.content_type == 'application/x-www-form-urlencoded':
         body = request.POST.mixed()
 
-    elif request.content_type and request.content_type != 'application/json':
+    elif (request.content_type and
+          not re.match('^application/(.*?)json$', str(request.content_type))):
```
gabisurita: It's expecting ASCII. This is just there because otherwise we cant [mock it](https://github.com/Cornices/cornice/blob/master/tests/test_util.py#L12). But I'll change it to `six.binary_type`.

leplatrem: Side note, Pyramid has compat module to avoid using `six`:
http://docs.pylonsproject.org/projects/pyramid/en/latest/api/compat.html

cornice/validators/__init__.py
```diff
@@ -33,7 +34,8 @@ def extract_cstruct(request):
     if request.content_type == 'application/x-www-form-urlencoded':
         body = request.POST.mixed()
 
-    elif request.content_type and request.content_type != 'application/json':
+    elif (request.content_type and
```
leplatrem: I would like to suggest an intermediary variable here (like `is_json`) to avoid multiple line `if`.

cornice/validators/__init__.py
```diff
@@ -1,6 +1,7 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
+import re
```
leplatrem: nitpick: add a blank line between standard lib import and vendors

leplatrem:
Thanks @gabisurita! Well done!

PR #434, opened 2017-01-18
Add support for arrays on request body top level (fixes #433)

### LINKED ISSUE #433: Allow json body to be a sequence with colander_validator
Right now JSON validators just accepts objects as JSON bodies. Is there a reason to not accept Arrays as well?

From https://github.com/Kinto/kinto/issues/880

gabisurita:
r? @glasserc

### DIFF
--- a/cornice/validators/__init__.py
+++ b/cornice/validators/__init__.py
@@ -48,7 +48,7 @@ def extract_cstruct(request):
             else:
                 if not hasattr(body, 'items') and not isinstance(body, list):
                     request.errors.add('body', '',
-                                       'Should be a JSON object or array')
+                                       'Should be a JSON object or an array')
                     return {}
         else:
             body = {}
--- a/cornice/validators/__init__.py
+++ b/cornice/validators/__init__.py
@@ -46,8 +46,9 @@ def extract_cstruct(request):
                 request.errors.add('body', '', 'Invalid JSON: %s' % e)
                 return {}
             else:
-                if not hasattr(body, 'items'):
-                    request.errors.add('body', '', 'Should be a JSON object')
+                if not hasattr(body, 'items') and not isinstance(body, list):
+                    request.errors.add('body', '',
+                                       'Should be a JSON object or array')
                     return {}
         else:
             body = {}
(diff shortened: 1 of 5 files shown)


cornice/validators/__init__.py
```diff
@@ -46,8 +46,9 @@ def extract_cstruct(request):
                 request.errors.add('body', '', 'Invalid JSON: %s' % e)
                 return {}
             else:
-                if not hasattr(body, 'items'):
-                    request.errors.add('body', '', 'Should be a JSON object')
+                if not hasattr(body, 'items') and not isinstance(body, list):
+                    request.errors.add('body', '',
+                                       'Should be a JSON object or array')
```
Natim: Should be a JSON object or an array

cornice/validators/_colander.py
```diff
@@ -32,12 +32,11 @@ def body_validator(request, schema=None, deserializer=None, **kwargs):
     class RequestSchema(colander.MappingSchema):
         body = _ensure_instantiated(schema)
 
-        def deserialize(self, cstruct=colander.null):
-            appstruct = super(RequestSchema, self).deserialize(cstruct)
-            return appstruct['body']
-
     validator(request, RequestSchema(), deserializer, **kwargs)
 
+    if 'body' in request.validated:
```
gabisurita: This change is needed so we don't break the `request.validated.update(deserialized)` call on the base validator.
https://github.com/gabisurita/cornice/blob/14e783a9189c3557b388cc5183ab4490b9bea491/cornice/validators/_colander.py#L80

leplatrem: I didn't get that part :( When is it relevant to have this test? With the above RequestSchema it looks like `body` is mandatory no?

gabisurita: It is mandatory and it will always be present IF the validation [was successful](https://github.com/gabisurita/cornice/blob/14e783a9189c3557b388cc5183ab4490b9bea491/cornice/validators/_colander.py#L80). But it can be missing if the the deserialization fails.

We could also replace it for a single line like `request.validated = request.validated.get('body', {})`

PR #437, opened 2017-01-23
Add py35 and py36 to travis-ci

j-carl:


Natim:
Unfortunately it is [not that easy](https://travis-ci.org/Cornices/cornice/jobs/194584802), you would need to define a matrix. See how we did it there: 

https://github.com/Kinto/kinto/blob/master/.travis.yml#L28-L32

Issue #452, opened 2017-08-04
[need-help] Align Cornice with Pyramid ACL pattern


leplatrem:
Cornice comes in the way of several developers when it comes to dynamic ACLs, especially for resources.

The following is a very elegant pattern [proposed](https://github.com/Cornices/cornice/issues/120#issuecomment-284136414) by @DaineDanielson

```py
@resource(collection_path='/pages', path='/pages/{id}')
class PageResource:
    def __init__(self, request):
        self.id = uuid4()
        self.creator_id = request.user_id

    def __acl__(self):
        return [(Allow, Authenticated, 'add'),
                (Allow, self.creator_id, 'read')]

    @view(permission='add')
    def collection_post(self):
        self.request.response.status_code = 201
        return self.id
     
    @view(permission='read')
    def get(self):
        # ... retrieve page here ...
        return page
```

I'm not super available for this, so if someone is interested to lead that work, I would be happy to review, merge contributions and give admin privileges on the repo :)

wjehenddher:
It seems like using the `factory` argument, the scenario above can be achieved almost out of the box. 

```python
@resource(collection_path='/pages', path='/pages/{id}',

    factory=__name__ + ".PageResource"

)
class PageResource:
    def __init__(
        self, 
        request,
        # to prevent mismatched signature from cornice.service.Service.decorate_view.wrapper 
        #        ob = args['klass'](**params)  
        *args,
        **kwargs
    ):
        self.id = uuid4()
        self.creator_id = request.user_id

    def __acl__(self):
        return [(Allow, Authenticated, 'add'),
                (Allow, self.creator_id, 'read')]

    @view(permission='add')
    def collection_post(self):
        self.request.response.status_code = 201
        return self.id
     
    @view(permission='read')
    def get(self):
        # ... retrieve page here ...
        return page
```

Should the new implementation automatically wire `factory` to the `@resource` decorated class if no `klass` is explicitly declared?

Thoughts?

wjehenddher:
@leplatrem et al,

I pinged Pyramid IRC and got very good info from **raydeo**, one of the moderators.

In Pyramid theory, `PageResource.__acl__()` should be agnostic of `PageResource.id` and hence agnostic of the actual `request`. Instead, `PageResource.__acl__` should return the full list of ACEs ignoring current user and/or current request.

[...]

PR #456, opened 2017-09-13
Aligning Cornice with Pyramid ACL pattern by auto-wiring factory to @resource/@service decorated class

wjehenddher:
#452 

Dynamic `__acl__` by making the `@resource`/`@service` decorated class its own factory so its `__acl__` is used for authz

### DIFF
--- a/cornice/resource.py
+++ b/cornice/resource.py
@@ -7,7 +7,6 @@ import warnings
 import venusian
 
 from cornice import Service
-from pyramid.exceptions import ConfigurationError
 
 
 def resource(depth=2, **kw):
@@ -87,9 +86,6 @@ def add_resource(klass, depth=1, **kw):
         # is explicitly declared.
         if 'factory' not in kw:
             service_args['factory'] = klass
-
-        if 'acl' in service_args:
-            raise ConfigurationError("'acl' is not supported")
 
         # create service
         service_name = (service_args.pop('name', None) or
--- a/cornice/resource.py
+++ b/cornice/resource.py
@@ -82,6 +82,13 @@ def add_resource(klass, depth=1, **kw):
             elif k not in service_args:
                 service_args[k] = kw[k]
 
+        # auto-wire klass as its own view factory, unless one
+        # is explicitly declared. Make sure acl is thrown away
+        if 'factory' not in kw:
+            service_args['factory'] = klass
+            if 'acl' in service_args:
+                del(service_args['acl'])
+
         # create service
         service_name = (service_args.pop('name', None) or
                         klass.__name__.lower())
--- a/cornice/resource.py
+++ b/cornice/resource.py
@@ -7,6 +7,7 @@ import warnings
 import venusian
 
 from cornice import Service
+from pyramid.exceptions import ConfigurationError
 
 
 def resource(depth=2, **kw):
@@ -83,11 +84,12 @@ def add_resource(klass, depth=1, **kw):
                 service_args[k] = kw[k]
 
         # auto-wire klass as its own view factory, unless one
-        # is explicitly declared. Make sure acl is thrown away
+        # is explicitly declared.
         if 'factory' not in kw:
             service_args['factory'] = klass
-            if 'acl' in service_args:
-                del(service_args['acl'])
+
+        if 'acl' in service_args:
+            raise ConfigurationError("'acl' is not supported")
 
         # create service
         service_name = (service_args.pop('name', None) or
--- a/cornice/service.py
+++ b/cornice/service.py
@@ -81,15 +81,11 @@ class Service(object):
     :param factory:
         A factory returning callables which return boolean values.  The
         callables take the request as their first argument and return boolean
-        values.  This param is exclusive with the 'acl' one.
-
-    :param acl:
-        A callable accepting a request and returning a list of ACLs.
-        Exclusive with the 'factory' option.
+        values.
 
     :param permission:
         As for ``pyramid.config.Configurator.add_view()``.
-        Note: `acl` and `permission` can also be applied
+        Note: `permission` can also be applied
         to instance method decorators such as :meth:`~get` and :meth:`~put`.
 
     :param klass:
@@ -188,7 +184,7 @@ class Service(object):
                 setattr(self, key, value)
 
         if hasattr(self, 'acl'):
-            raise ConfigurationError("Cannot specify 'acl'")
+            raise ConfigurationError("'acl' is not supported")
 
         # instantiate some variables we use to keep track of what's defined for
         # this service.
@@ -472,10 +468,10 @@ def decorate_view(view, args, method, route_args={}):
         if 'klass' in args and not callable(view):
             # XXX: given that request.context exists and root-factory
             # only expects request param, having params seems unnecessary
+            # ob = args['klass'](request)
             params = dict(request=request)
             if 'factory' in route_args and 'acl' not in route_args:
                 params['context'] = request.context
-            # ob = args['klass'](request)
             ob = args['klass'](**params)
             if is_string(view):
                 view_ = getattr(ob, view.lower())
--- a/cornice/service.py
+++ b/cornice/service.py
@@ -465,9 +465,11 @@ def decorate_view(view, args, method, route_args={}):
         ob = None
         view_ = view
         if 'klass' in args and not callable(view):
+            #XXX: given that request.context exists and root-factory only expects request param, having params seems unnecessary
             params = dict(request=request)
             if 'factory' in route_args and 'acl' not in route_args:
                 params['context'] = request.context
+            #ob = args['klass'](request)
             ob = args['klass'](**params)
             if is_string(view):
                 view_ = getattr(ob, view.lower())
--- a/cornice/service.py
+++ b/cornice/service.py
@@ -465,11 +465,12 @@ def decorate_view(view, args, method, route_args={}):
         ob = None
         view_ = view
         if 'klass' in args and not callable(view):
-            #XXX: given that request.context exists and root-factory only expects request param, having params seems unnecessary
+            # XXX: given that request.context exists and root-factory
+            # only expects request param, having params seems unnecessary
             params = dict(request=request)
             if 'factory' in route_args and 'acl' not in route_args:
                 params['context'] = request.context
-            #ob = args['klass'](request)
+            # ob = args['klass'](request)
             ob = args['klass'](**params)
             if is_string(view):
                 view_ = getattr(ob, view.lower())
--- a/cornice/service.py
+++ b/cornice/service.py
@@ -186,6 +186,9 @@ class Service(object):
         if hasattr(self, 'acl'):
             raise ConfigurationError("'acl' is not supported")
 
+        if hasattr(self, 'traverse'):
+            raise ConfigurationError("'traverse' is not supported")
+
         # instantiate some variables we use to keep track of what's defined for
         # this service.
         self.defined_methods = []
@@ -275,10 +278,6 @@ class Service(object):
             view = _UnboundView(kwargs['klass'], view)
 
         args = self.get_arguments(kwargs)
-
-        # These attributes were used in views in Cornice < 1.0.
-        deprecated_attrs = ('acl', 'traverse')
-        args = {k: v for k, v in args.items() if k not in deprecated_attrs}
 
         # remove 'factory' if present,
         # it's not a valid pyramid view param
@@ -470,7 +469,7 @@ def decorate_view(view, args, method, route_args={}):
             # only expects request param, having params seems unnecessary
             # ob = args['klass'](request)
             params = dict(request=request)
-            if 'factory' in route_args and 'acl' not in route_args:
+            if 'factory' in route_args:
                 params['context'] = request.context
             ob = args['klass'](**params)
             if is_string(view):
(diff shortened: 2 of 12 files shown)


cornice/service.py
```diff
@@ -465,9 +465,11 @@ def wrapper(request):
         ob = None
         view_ = view
         if 'klass' in args and not callable(view):
+            #XXX: given that request.context exists and root-factory only expects request param, having params seems unnecessary
```
wjehenddher: If this was done, `__init__()` would not need to be overloaded by decorated class. For the time being, it's left as is so decorated classes must declare `context` as optional param.

In current test-cases, `ThingImp` and `UserImp` have `context` as optional param in their `__init__` methods.

See:

https://github.com/Cornices/cornice/blob/76b0390a333d59b0977646aacb76809c7829d3ad/tests/test_imperative_resource.py#L30

https://github.com/Cornices/cornice/blob/76b0390a333d59b0977646aacb76809c7829d3ad/tests/test_imperative_resource.py#L40

cornice/service.py
```diff
@@ -465,9 +465,11 @@ def wrapper(request):
...
             params = dict(request=request)
             if 'factory' in route_args and 'acl' not in route_args:
                 params['context'] = request.context
+            #ob = args['klass'](request)
```
wjehenddher: If `__init__` didn't need overloading.

cornice/resource.py
```diff
@@ -82,6 +82,13 @@ class User(object):
             elif k not in service_args:
                 service_args[k] = kw[k]
 
+        # auto-wire klass as its own view factory, unless one
+        # is explicitly declared. Make sure acl is thrown away
+        if 'factory' not in kw:
+            service_args['factory'] = klass
+            if 'acl' in service_args:
+                del(service_args['acl'])
```
leplatrem: nitpick: superfluous parenthesis

tests/test_resource.py
```diff
@@ -151,6 +155,7 @@ def test_explicit_service_name(self):
         route_url = testing.DummyRequest().route_url
         self.assert_(route_url('user_service', id=42))  # service must exist
 
+    @skip('deprecated collection_acl param')
```
leplatrem: We should remove the test then IMO

cornice/resource.py
```diff
@@ -82,6 +82,13 @@ class User(object):
             elif k not in service_args:
                 service_args[k] = kw[k]
 
+        # auto-wire klass as its own view factory, unless one
+        # is explicitly declared. Make sure acl is thrown away
+        if 'factory' not in kw:
+            service_args['factory'] = klass
+            if 'acl' in service_args:
```
leplatrem: Shouldn't we raise here: `acl is not supported...` ?

tests/test_resource.py
```diff
@@ -162,15 +167,42 @@ class List(object):
...
+    @mock.patch('cornice.resource.Service')
+    def test_factory_is_autowired(self, mocked_service):
+        @resource(collection_path='/list', path='/list/{id}', name='list')
+        class List(object):
+            pass
+        factory_args = [kw.get('factory') for _, kw in mocked_service.call_args_list]
+        self.assertEqual([List, List], factory_args)
+
+    @mock.patch('cornice.resource.Service')
+    def test_acl_is_discarded(self, mocked_service):
```
leplatrem: As seen earlier, instead of silently discard it, maybe we should raise explicitly?

cornice/service.py
```diff
@@ -465,9 +465,12 @@ def wrapper(request):
         ob = None
         view_ = view
         if 'klass' in args and not callable(view):
+            # XXX: given that request.context exists and root-factory
```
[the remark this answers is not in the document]
wjehenddher: BTW, please take a look at the "outdated" comments I made.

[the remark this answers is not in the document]
wjehenddher: I ran into issues with the `__init__(self, request)`:

https://github.com/Cornices/cornice/blob/76b0390a333d59b0977646aacb76809c7829d3ad/tests/test_service.py#L506-#L530

These tests assert `context`.

The problem stems from `DummyAPI` trying to capture `context` when `__init__()` is being called.

https://github.com/Cornices/cornice/blob/76b0390a333d59b0977646aacb76809c7829d3ad/tests/test_service.py#L26-#L37

Naively I thought that `request.context` could be used instead. It **cannot** because the *context* per se is being *constructed* since we are in the `__init__` of the `factory`.

> No matter which view calling convention is used, the view code always has access to the context via request.context.

See https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/views.html#alternate-view-callable-argument-calling-conventions

```python
@resource(collection_path='/pets', path='/pets/{id}')
class DummyAPI(object):
    last_request = None
    last_context = None


    def __init__(self, request, context=None):
        DummyAPI.last_request = request
        DummyAPI.last_context = request.context ### Naive attempt


    def collection_get(self):
        return ['douggy', 'rusty']
```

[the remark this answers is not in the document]
wjehenddher: I don't think we want to remove the 2 failing assertions dealing with `DummyAPI.last_context`.

docs/source/upgrading.rst
```diff
@@ -1,6 +1,40 @@
 Upgrading
 #########
 
+2.X to 3.X
+==========
+
+`acl` parameter is deprecated and a class decorated with `@resource`
```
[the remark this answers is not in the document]
leplatrem: Also, since we raise a `ConfigurationError`, we should probably say `parameter is not supported anymore` instead of `deprecated`

cornice/service.py
```diff
@@ -281,9 +277,14 @@ def add_view(self, method, view, **kwargs):
         args = self.get_arguments(kwargs)
 
         # These attributes were used in views in Cornice < 1.0.
-        deprecated_attrs = ('acl', 'factory', 'traverse')
+        deprecated_attrs = ('acl', 'traverse')
```
leplatrem: I think `acl` does not have to be here since we raise if it is specified. We could do the same as above with `self.traverse`, since we bump a major

cornice/service.py
```diff
@@ -465,6 +466,9 @@ def wrapper(request):
         ob = None
         view_ = view
         if 'klass' in args and not callable(view):
+            # XXX: given that request.context exists and root-factory
+            # only expects request param, having params seems unnecessary
+            # ob = args['klass'](request)
             params = dict(request=request)
             if 'factory' in route_args and 'acl' not in route_args:
```
leplatrem: Can `'acl'` still be in route args?

cornice/service.py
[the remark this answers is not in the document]
leplatrem: From what I see in the tests and code, it looks like you managed to do it, didn't you?

cornice/resource.py
```diff
@@ -82,6 +83,14 @@ class User(object):
             elif k not in service_args:
                 service_args[k] = kw[k]
 
+        # auto-wire klass as its own view factory, unless one
+        # is explicitly declared.
+        if 'factory' not in kw:
+            service_args['factory'] = klass
+
+        if 'acl' in service_args:
+            raise ConfigurationError("'acl' is not supported")
```
leplatrem: Is this necessary, since service already raises?

docs/source/upgrading.rst
```diff
@@ -1,6 +1,40 @@
...
+or `@service` becomes its own `route factory
+<http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/urldispatch.html#route-factories>`_ and
+it may take advantage of `Pyramid ACL
+<https://docs.pylonsproject.org/projects/pyramid/en/latest/narr/security.html#assigning-acls-to-your-resource-objects>`_.
+
+Before:
+
+.. code-block:: python
+
+    @resource(path='/users')
```
leplatrem: Here I think it would be relevant to have  `acl=[(Allow, Everyone, 'view')]` no?

cornice/service.py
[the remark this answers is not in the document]
wjehenddher: No. I couldn't because of the `DummyAPI.last_context` test-cases I cited 😢 

I didn't want to remove these tests either because they are verifying what the pyramid `context` is in each case.  See [pyramid context](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/request.html#pyramid.request.Request.context)

Ideally, I would have liked to rewrite the tests and assert that the `context` is an instance created by the *auto-wired* `factory` but I didn't come up with a clever solution to check what the `context` was in each case (as the tests do) given that the `context` is itself when the `__init__` method is called.

I didn't want to remove the `# XXX:` I added in case I or anyone else feels like tacking this *naunce*.

docs/source/upgrading.rst
```diff
@@ -1,6 +1,40 @@
...
+    class User(object):
+
+        def __init__(self, request):
+            self.request = request
+            self.user = context
```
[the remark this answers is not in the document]
wjehenddher: Whoops! 🤦‍♂️

cornice/service.py
wjehenddher: Will `raise ConfigurationError()` for `traverse` as well and upgrade rSTs.

wjehenddher:
Hi @leplatrem @tarekziade 

Please let me know if there is anything else I should do/change.

From my point of view, I think the feature is complete 🎉

wjehenddher:
Thank you @leplatrem!

Should we give @tarekziade few more days to review the changes?

Will I get a "Contributor" badge? I would like that very much 😃

Natim:
> Will I get a "Contributor" badge

Apparently you have 👍 congrats !!

wjehenddher:
Yay!!!



On Sep 26, 2017, at 9:10 AM, Rémy HUBSCHER <notifications@github.com<mailto:notifications@github.com>> wrote:


Will I get a "Contributor" badge

Apparently you have 👍 congrats !!

joesteeve:
Sorry that I am late to the party. 

When a `route` has a `route_factory`, pyramid will use it to create a `context`. When there is no `route_factory` attached with a `route`, pyramid will use the `root factory` to create a `root` object, and use it as the context. The ACLAuthorisationPolicy looks for `__acl__` on the context and its lineage.

Given the above, by making the resource/service decorators treat CBVs as `route_factories`, we are effectively ignoring the `root-factory`. It is a common pattern for URL dispatch applications to construct a root object with the ACL for that particular user, and use it for all routes. This change breaks that pattern.

leplatrem:
Thanks @joesteeve for your feedback!

Do you have a proposition that would both satisfy developer experience and honour Pyramid patterns?

joesteeve:
Hello @leplatrem ,

IMHO, explicit setting of a route-factory in the `@resource` decorator will be most pyramid friendly. It'll allow the developer to attach an explicit context to the resource routes. The context could have an `__acl__` attribute. Or, if it wants inherited permissions, it could become `location aware` by becoming a child of the `root`.

It could look like:

```python
@resource(path='/users', factory=custom_factory,
    collection_factory=custom_collection_factory)
class User(object):
    def __init__(self, request, context=None):
        self.request = request
        self.context = context
```

@joshuaimmanuel . Your thoughts would be welcome too.

joshuaimmanuel:
Here in this implementation, if the `factory` argument is not specified in the `@resource` decorator, they are making the current decorated class as *route factory*. And this **breaks** the pyramid pattern. 

Instead of implicitly considering the decorated class as *route factory* if `factory` argument is not supplied, I think we should make it explicit. I mean supply a parameter to make the decorated class as *route factory*, and if `factory` parameter is specified use that as *route factory*, and if `factory` parameter is not specified use the *default root factory*

I think, this way we can achieve what we want without breaking the pyramid pattern.

joesteeve:
Ouch. I mixed up.  @joshuaimmanuel  is right. The `@resource` already retains the `route factory` behaviour. When a `route factory` is not given, we should use the `root` object as the context.

Instead if the developer wants to use the CBV as a `route factory`, he should do so explicitly.

wjehenddher:
Hello @joshuaimmanuel and @joesteeve.

#452 has a detailed discussion about this change.

Do you both think that this change should be reverted?

If memory doesn't fail me (it's being a year), the older behavior is somehow preserved when implementor explicitly specifies a `factory`.

joshuaimmanuel:
IMO, I think we should revert this change.

As I said earlier, if we still want to make the decorated class as *route factory* without breaking the pyramid pattern, we can do either of these:

1. Introduce another parameter in the `@resource` decorator **or**
2. Allow a marker for `factory` parameter .  Eg: `@resource(factory=cornice.I_AM_A_ROUTE_FACTORY, ...)` 

IMO, option (2) is cleaner approach (except for the marker name)

wjehenddher:
@joshuaimmanuel proposal 2 is a good idea:
> Allow a marker for factory parameter . Eg: `@resource(factory=cornice.I_AM_A_ROUTE_FACTORY, ...)` 

I would prefer a view param to turn **OFF** the auto-wiring though. IOW, `@resource(autowire_factory=False)`

@leplatrem @DaineDanielson?

grant-humphries:
Hi folks, 

It looks like this conversation has gone stale, but I am having similar issues to those raised in this thread as I am upgrading from version 1.0.0 to 3.5.1. We have have a bunch of existing services that are expecting the Pyramid root factory to be used as the factory for cornice resources when no factory parameter is passed to the decorator.  

It seems like too much time has passed for reverting these changes to be an option, but I'm just looking for something to be present in the API that either allows me to set a default factory for all resources or a setting that indicates that the Pyramid root factory should be the default factory for all resources.

PR #464, opened 2017-10-25
Convert a bare except to make flake8 E722 pass.

bowlofeggs:


Natim:
Thanks

PR #465, opened 2017-10-30
Add validators for querystring, path, and headers.

bowlofeggs:

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,7 +5,10 @@ CHANGELOG
 3.1.0 (unreleased)
 ==================
 
-- Nothing changed yet.
+- In addition to the ``colander_body_validator``, there are now three more similar validators:
+  ``colander_headers_validator``, ``colander_path_validator``, and
+  ``colander_querystring_validator``.
+- six is now a required dependency.
 
 
 3.0.0 (2017-10-20)
--- a/cornice/validators/__init__.py
+++ b/cornice/validators/__init__.py
@@ -6,11 +6,17 @@ import re
 from webob.multidict import MultiDict
 from cornice.validators._colander import (
     validator as colander_validator,
-    body_validator as colander_body_validator)
+    body_validator as colander_body_validator,
+    headers_validator as colander_headers_validator,
+    path_validator as colander_path_validator,
+    querystring_validator as colander_querystring_validator)
 
 
 __all__ = ['colander_validator',
            'colander_body_validator',
+           'colander_headers_validator',
+           'colander_path_validator',
+           'colander_querystring_validator',
            'extract_cstruct',
            'DEFAULT_VALIDATORS',
            'DEFAULT_FILTERS']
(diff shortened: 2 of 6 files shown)


bowlofeggs:
I'll take a look at the test failures soon, but it might take me a few days. I'm busy updating one of my apps to Cornice 3 from Cornice 1 (whew!)

bowlofeggs:
@leplatrem Thanks for the approval!

What do you think about using ```six``` to resolve the metaclass differences between Python 2 and 3 for this patch?

Here's a change I'm applying to my patch that I think will get it to pass:

```
diff --git a/cornice/validators/_colander.py b/cornice/validators/_colander.py                                        
index 85fdb5f..0aa2305 100644                              
--- a/cornice/validators/_colander.py                      
+++ b/cornice/validators/_colander.py                      
@@ -5,6 +5,8 @@              
 import inspect              
 import warnings             
                             
+from six import with_metaclass                            
+                            
                             
 def _generate_colander_validator(location):               
     """                     
@@ -60,9 +62,10 @@ def _generate_colander_validator(location):                                                        
                 class_attrs[location] = _ensure_instantiated(schema)                                                 
                 return type(name, bases, class_attrs)     
                             
-        class RequestSchema(colander.MappingSchema):      
+        class RequestSchema(with_metaclass(               
+                RequestSchemaMeta, colander.MappingSchema)):                                                         
             """A schema to validate the request's location attributes."""                                            
-            __metaclass__ = RequestSchemaMeta             
+            pass            
                             
         validator(request, RequestSchema(), deserializer, **kwargs)                                                  
         request.validated = request.validated.get(location, {})                                                      
diff --git a/setup.py b/setup.py                           
index d406235..9af21d1 100755                              
--- a/setup.py               
+++ b/setup.py               
@@ -9,7 +9,7 @@ with open(os.path.join(here, 'README.rst')) as f:                                                     
 with open(os.path.join(here, 'CHANGES.txt')) as f:        
     CHANGES = f.read()      
                             
-requires = ['pyramid>=1.7',  'simplejson']                
+requires = ['pyramid>=1.7',  'simplejson', 'six']         
                             
 entry_points = ""           
 package_data = {}
```

leplatrem:
> What do you think about using six to resolve the metaclass differences between Python 2 and 3 for this patch?

I don't have too much bandwidth lately, but otherwise I would be tempted to drop Python2 support :) 

Regarding `six` I have nothing against it, but if it's just for one helper, I believe we can put it inline in our utils instead of adding a new dependency. Looks it's rather small:
https://github.com/benjaminp/six/blob/63b2661b94463b59752396a3562c9e5ebe672f97/six.py#L819-L832

leplatrem:
@Natim what do you think about inlining the metaclass helper instead of adding dependency to `six`?

Natim:
What would be the benefit of doing that?

leplatrem:
> What would be the benefit of doing that?

Avoid introducing a new dependency that does a thousand things for a 8 LOC function

Natim:
I wouldn't be afraid about it, six is also most likely installed already.

PR #468, opened 2017-12-19
Require venusian in setup.py.

bowlofeggs:
Two modules import venusian, but it wasn't required in the
setup.py. This commit adds it there.


leplatrem:
Theorically we wouldn't need it because Pyramid already depends on it: 
https://github.com/Pylons/pyramid/blob/c9a91f4eaab7e5fc2d888ae94de202ddbfcdf96f/setup.py#L28

But since we import it explicitly I'm not against adding it to the dependencies here

Natim:
> @Natim what do you think?

Yes it makes a lot of sense to add it if we import it in the project.

PR #471, opened 2018-02-12
Make tutorial work for python3 users

rvandegrift:
I ran into some issues following the tutorial, here are the fixes I needed to complete it.

Thanks!

### DIFF
--- a/docs/source/tutorial.rst
+++ b/docs/source/tutorial.rst
@@ -41,21 +41,10 @@ We'll provide a single CLI client in Python, using Curses.
 Setting up the development environment
 --------------------------------------
 
-If you are using python 2, make sure you have **virtualenv** (see
-https://pypi.python.org/pypi/virtualenv).  If you use python 3, the
-relevant tool comes with python.
-
-Create a new directory and a virtualenv in it::
+To begin, create a new directory and environment::
 
     $ mkdir messaging
     $ cd messaging
-
-If you use python 2, use virtualenv to create your environment::
-
-    $ virtualenv --no-site-packages .
-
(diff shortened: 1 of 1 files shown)


leplatrem:
Thanks for taking the time to do this!

Since the tutorial is meant for new projects, I think we can fully switch to Python 3 and get rid of mentions of Python 2. What do you think?

docs/source/tutorial.rst
```diff
@@ -41,18 +41,31 @@ We'll provide a single CLI client in Python, using Curses.
 Setting up the development environment
 --------------------------------------
 
-Make sure you have **virtualenv** (see http://pypi.python.org/pypi/virtualenv).
+If you are using python 2, make sure you have **virtualenv** (see
```
leplatrem: I would write Python (capital)

docs/source/tutorial.rst
```diff
@@ -192,17 +204,19 @@ then use.
...
             user = {'name': name, 'token': _create_token()}
             request.validated['user'] = user
 
 
-When the validator finds errors, it adds them to the **request.errors**
-mapping, and that will return a 400 with the errors.
+The validators work by filling the **request.validated**
+dictionary. When the validator finds errors, it adds them to the
+**request.errors** dictionary, and that will return a 400 with the
+errors.
```
leplatrem: 👍

rvandegrift:
Sounds good - I've dropped the bit about v2.  That also removes the capitalization issue.

leplatrem:
Thanks @rvandegrift!

PR #475, opened 2018-03-17
Add Marshmallow validation support to Cornice

ergo:


ergo:
Thats weird, when I'm running tox locally I get 100% coverage and everything passes, any idea what might be different in relation to travis build?

leplatrem:
> Thats weird, when I'm running tox locally I get 100% coverage and everything passes, any idea what might be different in relation to travis build?

According to your comment at that part of the code, it depends on the marshmallow version (2.X or 3.X).  TravisCI runs `marshmallow==2.15.0`.

I guess that if you recreate your tox folder, you'll get the same coverage miss no?

I don't think we want to test marshmallow with several versions, so we should be fine by adding a `# pragma: no cover`

ergo:
Funnily enough https://coveralls.io/builds/16034156 - coveralls shows 100%, and yes even when I deleted `.tox` and run it again locally on fresh env I got 100% coverage.

Anyways, I've added `no cover` so it should pass now.

leplatrem:
Thanks!

It looks good :) 

Would you mind adding a few lines to mention this new stuff in the docs? And a few lines in the changelog too?
https://cornice.readthedocs.io/en/latest/schema.html

ergo:
Sure, I can add docs and changelog, first I wanted to get confirmation everything is in order to avoid double work on docs.

ergo:
@leplatrem The documentation has landed.

Issue #440, opened 2017-02-28
Cornice should not perform a commit on Configurator


leplatrem:
See https://github.com/Cornices/cornice/issues/151#issuecomment-282786619

ergo:
Should be fixed in by https://github.com/Cornices/cornice/pull/476

PR #476, opened 2018-03-24
configurator: cornice should not commit configuration ever

ergo:
Fixes https://github.com/Cornices/cornice/issues/440, by removing commits and fixing tests with conflicting declarations.

### DIFF
--- a/cornice/pyramidhook.py
+++ b/cornice/pyramidhook.py
@@ -253,15 +253,12 @@ def register_service_views(config, service):
             config.add_view(view=decorated_view, route_name=route_name,
                             **args)
 
-        config.commit()
-
     if service.definitions:
         # Add the fallback view last
         config.add_view(view=get_fallback_view(service),
                         route_name=route_name,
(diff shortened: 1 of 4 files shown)


ergo:
Basicly if you do:

    @view_config('foo', renderer='foo')
    @view_config('foo', renderer='bar')
    def bar...

How pyramid is supposed to understand what renderer it should use? It need to distinguish request based on some condition. This is exactly why the test failed.

leplatrem:
> How pyramid is supposed to understand what renderer it should use? It need to distinguish request based on some condition. This is exactly why the test failed.

Yes I get that, but I now wonder how it worked when we were doing the `config.commit()1`

ergo:
I don't know, it should not be working to my knowledge. Maybe something swallowed the exception? (Btw. I'm online on #services-dev if you want to chat.

leplatrem:
Fine, we should not be committing anyway, so let's trust the tests on this one :) 

Thanks!

Issue #137, opened 2013-04-13
Not making the path mandatory for Service and let the user register the route 


rach:
Because the root is automatically set via Cornice.

I'm loosing the fine control of being able to add the route myself abd set a Pregenerator or Predicate 

I'm not sure how big the change will be but the idea is only adding the route if a path is defined. It will fail if service is defined without path and the user didn't set a route for the service name.

https://github.com/mozilla-services/cornice/blob/master/cornice/pyramidhook.py#L180

Why?

Like I said previously, It seems that I'm not able to set a custom predicate or pregenerator for the service route.

My use case is to be able to have service path which match only against a specific subdomain.

I'm maybe wrong then let know how I could work around my issues.

doc:
[pregenerator] http://docs.pylonsproject.org/projects/pyramid/en/latest/api/interfaces.html#pyramid.interfaces.IRoutePregenerator

[predicate] http://docs.pylonsproject.org/projects/pyramid/en/latest/glossary.html#term-route-predicate

Issue #473, opened 2018-03-01
route_url() and pregenerator


jenstroeger:
I noticed that Pyramid allows to pass a [pregenerator](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/interfaces.html#pyramid.interfaces.IRoutePregenerator) to [`request.add_route()`](https://docs.pylonsproject.org/projects/pyramid/en/latest/api/config.html#pyramid.config.Configurator.add_route), which is then used by `request.route_url()` to touch up its arguments.

I don’t see this parameter for [`Service()`](https://cornice.readthedocs.io/en/latest/api.html#cornice.service.Service) though. What’s the recommended way of adding such a pregenerator?

leplatrem:
It doesn't seem to be possible to pass custom route arguments currently.

But we could add a service attribute (like `predicate`) here for example:
https://github.com/Cornices/cornice/blob/494f35f821f9456b80cca467ffb502d248f25c19/cornice/pyramidhook.py#L183-L201

PR #477, opened 2018-03-27
allow services to reuse existing pyramid routes instead ofcreating new ones every time

ergo:
Services: allow services to reuse existing pyramid routes instead ofcreating new ones every time

I believe this will fix: 

https://github.com/Cornices/cornice/issues/137 and https://github.com/Cornices/cornice/issues/473 among things.

### DIFF
--- a/cornice/resource.py
+++ b/cornice/resource.py
@@ -62,6 +62,10 @@ def add_resource(klass, depth=1, **kw):
 
     services = {}
 
+    if (('collection_pyramid_route' in kw or 'pyramid_route' in kw) and
+            ('collection_path' in kw or 'path' in kw)):
+        raise ValueError('You use either paths or route names, not both')
+
     if 'collection_path' in kw:
         if kw['collection_path'] == kw['path']:
             msg = "Warning: collection_path and path are not distinct."
@@ -70,6 +74,14 @@ def add_resource(klass, depth=1, **kw):
         prefixes = ('', 'collection_')
     else:
         prefixes = ('',)
+
+    if 'collection_pyramid_route' in kw:
+        if kw['collection_pyramid_route'] == kw['pyramid_route']:
+            msg = "Warning: collection_pyramid_route and " \
+                  "pyramid_route are not distinct."
+            warnings.warn(msg)
+
+        prefixes = ('', 'collection_')
 
     for prefix in prefixes:
 
--- a/cornice/service.py
+++ b/cornice/service.py
@@ -48,6 +48,9 @@ class Service(object):
     :param path:
         The path the service is available at. Should also be unique.
 
+    :param pyramid_route:
+        Use existing pyramid route instead of creating new one.
+
     :param renderer:
         The renderer that should be used by this service. Default value is
         'simplejson'.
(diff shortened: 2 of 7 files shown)


cornice/resource.py
```diff
@@ -71,6 +71,14 @@ class User(object):
...
         prefixes = ('',)
 
+    if 'collection_pyramid_route' in kw:
+        if kw['collection_pyramid_route'] == kw['pyramid_route']:
+            msg = "Warning: collection_pyramid_route and " \
+                  "pyramid_route are not distinct."
+            warnings.warn(msg)
+
+        prefixes = ('', 'collection_')
+
```
leplatrem: What happens when `*path` and `*route` are specified?

cornice/service.py
```diff
@@ -153,12 +156,19 @@ class Service(object):
...
-                 **kw):
+    def __init__(self, name, path=None, description=None, cors_policy=None,
+                 depth=1, pyramid_route=None, **kw):
         self.name = name
         self.path = path
+        self.pyramid_route = pyramid_route
+
+        if not self.path and not self.pyramid_route:
+            raise TypeError('You need to pass either path or '
+                            'pyramid_route arg')
```
leplatrem: nit: Maybe `ValueError` instead?

tests/test_resource.py
```diff
@@ -84,6 +84,17 @@ def __init__(self, request, context=None):
         msg = "Warning: collection_path and path are not distinct."
         mocked_warn.assert_called_with(msg)
 
+    @mock.patch('warnings.warn')
+    def test_routes_clash(self, mocked_warn):
+        @resource(collection_pyramid_route='/badthing/{id}',
+                  pyramid_route='/badthing/{id}', name='bad_thing_service')
```
leplatrem: those routes look like paths no?

tests/test_pyramidhook.py
```diff
@@ -268,6 +268,29 @@ def test_route_with_traverse_fails(self):
...
+        pyramid_route_service = Service(
+            name="SomeServiceWPyramidRoute", pyramid_route="pyramid_route")
+        pyramid_route_service.add_view(
+            'GET', lambda request: request.current_service.name)
+
+        register_service_views(self.config, pyramid_route_service)
+        self.app = TestApp(CatchErrors(self.config.make_wsgi_app()))
+
+    def test_service_routing(self):
+        self.app.get('/from_pyramid', status=200)
```
leplatrem: you could assert the route name in the response here maybe

tests/test_pyramidhook.py
```diff
@@ -268,6 +268,29 @@ def test_route_with_traverse_fails(self):
...
+
+    def setUp(self):
+        self.config = testing.setUp()
+        self.config.include("cornice")
+        self.config.add_route('pyramid_route', '/from_pyramid')
+        self.config.scan("tests.test_pyramidhook")
+        pyramid_route_service = Service(
+            name="SomeServiceWPyramidRoute", pyramid_route="pyramid_route")
+        pyramid_route_service.add_view(
+            'GET', lambda request: request.current_service.name)
```
leplatrem: you could use shorter variable names to prevent wrapping

cornice/service.py
```diff
@@ -153,12 +156,19 @@ class Service(object):
     list_arguments = ('validators', 'filters', 'cors_headers', 'cors_origins')
 
     def __repr__(self):
-        return u'<Service %s at %s>' % (self.name, self.path)
+        return u'<Service %s at %s>' % (
+            self.name, self.pyramid_route or self.path)
```
leplatrem: You could unit test this too :)

cornice/service.py
ergo: When argument is missing from call, python throws `TypeError`, so I followed that since proper arguments are missing. I can change to `ValueError` if you prefer that.

leplatrem: good point! don't change

cornice/service.py
ergo: Fixed.

cornice/resource.py
ergo: Fixed with test and exception

PR #482, opened 2018-04-07
Validation schemas should receive request object/get bound during validation

ergo:
Colander schemas that use deferred values should get bound to request object (and need to be bound in general) - otherwise the deferred values will never get resolved which will lead to wrong behavior.

For Marshmallow schemas it will also be beneficial when validators can access request object for pyramid registry, session and whatnot.


tests/validationapp.py
```diff
@@ -177,6 +179,23 @@ def post7(request):
     class SignupSchema(MappingSchema):
         username = SchemaNode(String())
 
+    @deferred
+    def deferred_missing(node, kw):
+        import random
+        return kw.get('missing_foo') or random.random()
```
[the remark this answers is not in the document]
ergo: The other test checks if the value changes per request (for example when you want to return a timestamp or UUID), so I picked this on purpose.

cornice/validators/_marshmallow.py
```diff
@@ -141,6 +143,9 @@ def validator(request, schema=None, deserializer=None, **kwargs):
     if schema is None:
         return
 
+    if not schema.context.get('request'):
+        schema.context['request'] = request
```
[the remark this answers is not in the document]
ergo: Sure

cornice/validators/_marshmallow.py
```diff
@@ -42,6 +42,8 @@ def _validator(request, schema=None, deserializer=None, **kwargs):
 
         class ValidatedField(marshmallow.fields.Field):
             def _deserialize(self, value, attr, data):
+                if not schema.context.get('request'):
+                    schema.context['request'] = request
```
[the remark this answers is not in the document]
leplatrem: Or maybe contribute a new `clone=False` parameter in colander?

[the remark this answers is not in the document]
ergo: I can ask @mmerickel about this but I think it needs to clone recursively everything when we bind.

tests/validationapp.py
```diff
@@ -305,6 +325,23 @@ def item(request):
...
+
+    class MNeedsContextSchema(marshmallow.Schema):
+        somefield = marshmallow.fields.Float(missing=lambda: random.random())
+        csrf_secret = marshmallow.fields.String()
+
+        @marshmallow.validates_schema
+        def validate_csrf_secret(self, data):
+            # simulate validation of session variables
+            if self.context['request'].get_csrf() != data.get('csrf_secret'):
+                raise marshmallow.ValidationError('Wrong token')
```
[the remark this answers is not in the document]
ergo: And its added to the docs

ergo:
@leplatrem I've removed the auto binding of colander - but left some tests for working with bound schemas. I think this should be handy.

When I first implemented Marshmallow support I thought following 1:1 colander behavior which seemed like a good idea, unfortunately that was a mistake, I don't think that we should ever be instantiating schemas on module/decorator level.

In case of Marshmallow and context binding that would lead to mutation of global schema object.
After chatting with colander maintainers they also said that it is required to clone all the nodes to avoid problems like that.

I've reworked docs, tests and validators to expect class only and instantiate it per request - only this ensures that we will be getting right results in both single/multi threaded environments.

I did it in a backwards compatible matter with a deprecation warning - however I strongly feel that I should drop the support for this completely, especially since cornice stable has support for Marshmallow for just 6 days, @leplatrem please share your thoughts on this.

```
    def rebinding_validator(request, **kwargs):
        kwargs['schema'] = NeedsBindingSchema().bind()
        return colander_body_validator(request, **kwargs)

    @bound.post(schema=NeedsBindingSchema().bind(), validators=(rebinding_validator,))
    def bound_post(request):
        return request.validated
```
This is what a "proper" way of working with marshmallow/colander would look like to have tools like `cornice_swagger` and application get proper values per every request. I think it is rather unwieldy API vs this (This would work properly for marshmallow already in this PR, but not with examples currently endorsed by cornice docs for colander schemas):

```
    @bound.post(schema=MarshmallowSchema, validators=(marshmallow_body_validator,))
    def bound_post(request):
        return request.validated
```

leplatrem:
I can't dig into the details as deep as you did, but as far as I understand, it kind of crosses what @dairiki did in https://github.com/Cornices/cornice/pull/415

In order to avoid going back and forth, I think I would need some help to make the right choice here :)

ergo:
@leplatrem currently in this PR this is completely related to Marshmallow - it will never work like colander - while bind() is slow as you mentioned, it also ensures everything is cloned correctly so everything will work as it should.

With Marshmallow we should never instantiate the schemas on module level because the objects are mutated.

cornice/validators/_marshmallow.py
```diff
@@ -163,3 +171,7 @@ def validator(request, schema=None, deserializer=None, **kwargs):
                 request.errors.add(location, location, details)
     else:
         request.validated.update(deserialized)
+
+
+def _instantiate_schema(schema):
```
leplatrem: Here I would raise an exception if `schema` is not a class. Just in case developers switch code from colander to marshmallow for example

ergo: Done

CHANGES.txt
```diff
@@ -5,7 +5,27 @@ CHANGELOG
...
+- Allow to reuse Pyramid existing routes in services (#477, thanks @ergo)
+
+**Documentation**
+
+- Make tutorial work for python3 users (#471, thanks @rvandegrift)
+
+**Internal changes**
+
+- Remove configuration commit (#476, thanks @ergo)
+- Remove mentions of ``text/json`` in tests and docs (#478, thanks @wjehenddher)
```
leplatrem: Looks like you duplicated the 3.2.0 changelog entry

CHANGES.txt
```diff
@@ -5,7 +5,27 @@ CHANGELOG
 3.3.0 (unreleased)
 ==================
 
-- Nothing changed yet.
+- Cornice now only accepts Marshmallow schemas instead of instances
+- Marshmallow schemas get request object added to their context
+  during validation
```
leplatrem: You can mention the pull-request number here

CHANGES.txt
ergo: Fixed, sorry

CHANGES.txt
ergo: Fixed

PR #485, opened 2018-04-16
Fix code block in tutorial.rst

karantan:
Fix for code block on the tutorial page:

screenshot 2018-04-16 16 19 30

url: https://cornice.readthedocs.io/en/latest/tutorial.html

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -1,28 +1,32 @@
 # flake8: noqa
 # -*- coding: utf-8 -*-
-import sys, os
+import datetime
+import os
+import sys
+import pkg_resources
 try:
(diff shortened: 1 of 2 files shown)


leplatrem:
> btw we should also update this: https://github.com/Cornices/cornice/blob/master/docs/source/conf.py#L23 to the latest version right?

OMG yes! I completely forgot!

karantan:
@leplatrem I've pushed the changes.

I've also modified a few different things:
- consistency in using `'` and not mixing `"` and `'`
- fix imports so that they are pep8 compliant (ref: https://www.python.org/dev/peps/pep-0008/#id23)
- add copyright year to be dynamic

PR #491, opened 2018-09-10
Fix DeprecationWarning for unrecognized backslash escapes

glasserc:
This became a warning as of Python 3.6 -- see
https://docs.python.org/3/whatsnew/3.6.html#deprecated-python-behavior.


leplatrem:
Only docstrings?

glasserc:
I believe this PR corrects every incorrect backslash escape throughout the cornice project.

PR #493, opened 2018-10-03
Remove WebOb deprecation warnings

abk-code:


leplatrem:
Thanks for starting this!

The code has to be changed slightly, the return type of the new method is not the same.

Also, don't forget CHANGELOG and CONTRIBUTORS file ;)

abk-code:
Few questions.
1. Checking some other code (nova, opestack) I think this may involve slightly bigger changes
than I thought. https://bit.ly/2P7yRtp - That seems direct API replacement.  But having said that
there seems some webob dependencies since older Py versions don't seem to have this API.
2. Do  I need to use best_match for older versions while use new API for newer versions?
3. How do I change the dependency? WebOb ==1.8.2 seems good enough though.

leplatrem:
I don't understand parts 1 and 2. Does `acceptable_offers()` rely on a minimum Python version?

> How do I change the dependency? WebOb ==1.8.2 seems good enough though.

You can add the minimum required version of Webob in `setup.py`

leplatrem:
@abk-code do you need any help for this?

abk-code:
Sure, I havn't put cycles in this. I will get back to this soon.

leplatrem:
@abk-code what's up? :)

abk-code:
Hey @leplatrem : Pls un-assign this from me. Not done much progress.

leplatrem:
Ok, no problem, don't worry :) Thanks for starting this!!

Issue #418, opened 2016-10-28
import json instead of using simplejson


Natim:
As far as I know `simplejson` is a drop-in replacement for json which ships in Python2.6+ and Python3.0+

Maybe we could import `json` directly instead?

Natim:
Actually no: ``json.dumps({"toto": 1}, use_decimal=True)``

    TypeError: __init__() got an unexpected keyword argument 'use_decimal'

leplatrem:
http://stackoverflow.com/a/1960649/141895

okin:
I tried to tackle this today because another package I try to port to Python 3 relies on cornice.

I found out that removing the use of `use_decimal` in `cornice.util` does not affect the tests at all.
Unfortunately this doesn't mean that simply switching out the modules is all it takes.

To further tackle this I implemented a custom encoder that decodes byte. Without this there would be some errors on Python 3.4 & 3.5 because this versions won't convert bytes to JSON (but 3.6+ will).
Now I have the tests running on Python 3 but I miss coverage on Python 2.

I'm not so sure how to tackle this. Maybe another test?
If someone wants to help the current (WIP) code is at [here](https://github.com/okin/cornice/tree/feature/drop_simplejson_replacing_with_customencoder).
Any help is appreciated!

leplatrem:
Thanks @okin for your investigations!

> I found out that removing the use of use_decimal in cornice.util does not affect the tests at all.

Ok that's bad, we should have a test that asserts the expected behavior. If we decide to drop it, we need to have a very clear upgrade path in the upgrade notes

> ...but I miss coverage on Python 2.

Adding another test is rarely a bad idea ;) 

But to be honest, I would be glad if we dropped Python 2 support. I would even say: «yeah, please do it and bump a major version of cornice».

SharpEdgeMarshall:
Any news here?

leplatrem:
Let's do it!

devilicecream:
I could take this one if it's ok for you guys!

devilicecream:
Added myself to the contributors, but besides that I think we can close this issue! @leplatrem

devilicecream:
@leplatrem just realized it may be useful to add this changes in the changelog file as well

PR #494, opened 2018-10-03
Reducing simplejson use

okin:
The aim of this is to make more use of the _json_ module from the standard library instead of the external _simplejson_ module.
This prepares closing #418 but because of problems mentioned in that issue this isn't yet a full switch.

It also enables tests with Python 3.7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,9 @@ matrix:
     - python: 3.6
       env:
         - TOX_ENV=py36
+    - python: "3.7-dev"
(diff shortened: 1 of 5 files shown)


leplatrem:
Cool, thanks!

Would you mind update the CHANGELOG and CONTRIBUTORS too please?

okin:
> Would you mind update the CHANGELOG and CONTRIBUTORS too please?
Not at all. Updates have been pushed :)

PR #499, opened 2018-12-07
Remove remaining usage of best_match()

leplatrem:


cornice/util.py
```diff
@@ -112,7 +112,7 @@ def match_accept_header(func, context, request):
     """
     acceptable = to_list(func(request))
     request.info['acceptable'] = acceptable
-    return request.accept.best_match(acceptable) is not None
+    return request.accept.acceptable_offers(acceptable)
```
glasserc: The function description says it returns True, not merely a truthy value.

PR #502, opened 2019-01-14
Translate Error description field when it is a TranslationString type

czekan:
When I add custom validation errors with `request.errors.add` method (eg. `request.errors.add('body', 'filed', _('Message'))` ) then cornice will not translate description for me. This behavior is not consistent with colander schema validation where messages are translated automatically with `request.localizer.translate` method.
This PR adds support for auto translation so it will not be necessary to call translation directly while passing description parameter to `request.errors.add`.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -5,7 +5,7 @@ CHANGELOG
 3.5.0 (unreleased)
 ==================
 
-- Nothing changed yet.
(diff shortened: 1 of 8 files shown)


czekan:
Hi @leplatrem !

I've just added docs for i18n in general. I'm not sure if that's what you wanted?
I've also updated CONTRIBUTORS and CHANGES.

Issue #503, opened 2019-01-14
Putting values in `request.validated` before calling colander_body_validator


czekan:
Let say we have a service:

```
service = Service(name='service', path='/service')

class ColSchema(MappingSchema):
    field = SchemaNode(String())

def custom_validator(request, **kwargs):
    # do some validation and put something in request.validated
    request.validated['something'] = 'something'

@service.post(schema=ColSchema(), validators=(custom_validator, colander_body_validator))
def service_handler(request):
    request.validated['something']   # <---- raises KeyError
    return {}
```

Getting `something` from request.validated raises KeyError because of the line [https://github.com/Cornices/cornice/blob/master/cornice/validators/_colander.py#L71](https://github.com/Cornices/cornice/blob/master/cornice/validators/_colander.py#L71)
which overwrites request.validated with request.validated['body'].

Fix for that would be just replacing line 71 in the cornice/validators/_colander.py:
```
        request.validated = request.validated.get(location, {})
```
with
```
        validated_location = request.validated.get(location, {})
        request.validated.update(validated_location)
        request.validated.pop(location, None)
```

but ... there is something that this change breaks.
What if someone sets schema which inherits from SequenceSchema rather than MappingSchema? Current code just rewrites request.validated (which is instantiated as dict in pyramidhook.py wrap_request()) with a list. I'm not sure if this should work like that or I'm just missing something?

leplatrem:
Since we have that here:

https://github.com/Cornices/cornice/blob/c8d3ad66ea551e507e12656d0a1bc0e5d407e205/cornice/pyramidhook.py#L148-L149

I would assume that `request.validated` should be a `dict`. 
Even the colander code should assume to find a dict there, no? 

We can also explicitly document that a SequenceSchema cannot be used.

leplatrem:
> This change would potentially break compatibility. If that's ok I can prepere PR with a solution.

Yes, we can bump a major version, no problem.

> Thing is that if someone would want to use SequenceSchema he can, but it needs to be used with colander_validator and than it will be deserialized to request.validated['body'] and request.validated will still remain a dict.

Makes sense! Yes!

PR #504, opened 2019-01-16
Feature/changes in colander validation

czekan:
@leplatrem  I thought it is better to explicitly raise an error with a meaningful message rather than just fail with ValueError (on call ``request.validated.update(some_non_empty_list)``) when schema is not an instance of ``colander.MappingSchema``. What do you think? #503


Rogalek:
I had the same issue in my app. I am going to borrow your fork after merging with current master and use in my code. Thanks.

leplatrem:
My bad, I really missed this PR :(

Rogalek:
so @leplatrem are you going to merge this?

leplatrem:
I released 4.0 with this change

PR #508, opened 2019-03-14
Update repository URL on docs

gforcada:
It points to a `mozilla-services` github organization which is no longer the case.

The same goes with the mailing list, but I could not find where it is now, if there is any :confused:


Natim:
Thank you.

PR #510, opened 2019-06-27
add validation support for multipart/form-data

birnbera:
The WebOb returned by `request.POST` supports boundary delimited `multipart/form-data` POSTs. This change merely calls the `mixed()` method on the `POST` body when the content-type is `multipart/form-data`, allowing validation to proceed in cases where the body is a combination of e.g. `json` and binary data.

### DIFF
--- a/cornice/validators/__init__.py
+++ b/cornice/validators/__init__.py
@@ -51,7 +51,7 @@ def extract_cstruct(request):
     """
     is_json = re.match('^application/(.*?)json$', str(request.content_type))
 
-    if request.content_type == 'application/x-www-form-urlencoded':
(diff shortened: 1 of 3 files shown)


leplatrem:
Thanks for this contribution!

Any chance you could add a small test please?

birnbera:
@leplatrem Thanks for reviewing this pull request! Just added a couple tests.

I had to make a new service endpoint to test multiple fields in a multipart request. I left a comment regarding why, but basically it's because, as far as I can tell, there isn't any native support in colander or marshmallow for a `MultiDict` when it actually contains multiple values for a given key. I've seen some recipes for how to subclass list types to facilitate deserialization, but it's up to the user to pass such as schema to the view in order to get them working.

birnbera:
Any chance this could be reviewed?

Issue #451, opened 2017-08-04
How to handle CORS preflight correctly for complex context factory?


jenstroeger:
Since there is no other place for general discussion on Cornice use (see https://github.com/Cornices/cornice/issues/447), I’m just going to raise my question here in the hopes for an answer.

Suppose this service instance:
```python
Service("foo", "/api/foo/{id}", factory=FooContext, cors_policy=…)
```
The `FooContext` then checks  the resources defined in the URL, e.g. the `{id}` above, and fails if there is a problem with the resources for the view.
```python
class FooContext(object):
    def __init__(self, request):
        user = … # Retrieve the user based on request’s Authorization field.
        foo = get_foo(user, request.matchdict["id"])
        if foo is None:
            request.errors.add("url", "id", "Invalid foo id")
            request.errors.status = 404 # Not Found
            return
```
I use this approach because I’d like to fail early, and I’d like to have all specified resources ready to go when I enter the view function.

The problem here arises with [CORS preflight OPTIONS](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request). In the above example, the resource acquisition in `FooContext` fails because the client (browser) does _not_ to send Authorization information during preflight. Thus, resource acquisition fails, OPTIONS fails, the entire request fails during preflight although the resources _are_ available!

A hacky solution would be:
```python
    def __init__(self, request):
        if request.method == 'OPTIONS':
            return
```
Setting `"credentials": True` in the [`cors_policy`](http://cornice.readthedocs.io/en/latest/api.html#cornice.service.Service) parameter for the Service() doesn’t work.

Another approach would be to define a whole new Service() instance with a different and empty context factory. But that seems cumbersome and too verbose.

What is the recommended approach to handle this?

Natim:
We had a similar problem with Kinto. You should definitely return a 200 for a CORS OPTIONS request anytime even for a 404, the CORS OPTIONS requests just give information that the request is allowed no information about the existence of the resource.

@leplatrem Do you remember how we support that with [KintoCore resources](https://github.com/Kinto/kinto/tree/master/kinto/core/resource)?

leplatrem:
Cornice adds the options view without factory (and no required permission):
https://github.com/Cornices/cornice/blob/239aeccff5f5ecc423b0c324d793705aaaae0d3e/cornice/pyramidhook.py#L178-L183

[...]

PR #513, opened 2019-08-22
Added wrapper for factory, ignoring OPTIONS requests when default options view used.

novikovav:
Basic example, demonstrating proposed way how to deal with factories being called during `OPTIONS` cors preflight request. See #451 https://github.com/Cornices/cornice/issues/451#issuecomment-523876630

I haven't added tests yet just because proposed approach might be incorrect and not worth putting efforts in it.

### DIFF
--- a/cornice/pyramidhook.py
+++ b/cornice/pyramidhook.py
@@ -158,6 +158,19 @@ def wrap_request(event):
         setattr(request, 'info', {})
 
 
+def _ignore_options_factory_wrapper(factory):
+    """Don't call factory for OPTIONS views because browser sends no
+    authorization header, which breaks some factories relying on user to be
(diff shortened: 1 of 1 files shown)


Natim:
I kinda like the approach. I don't know if there is a use case for which we would want to use the factory with OPTION.

@leplatrem I couldn't find how we manage to handle this with Kinto any idea?

jenstroeger:
@Natim, any progress on this?

Natim:
@jenstroeger do you know what kind of test we could write to make sure that this feature doesn't break in the future?

leplatrem:
@novikovav do you still need this? I'm trying to clean-up the state of the repo, please let me know if you want to keep this rolling!

Issue #514, opened 2019-09-02
marshmallow_validator seems to break with marshmallow version 3.0.1


mamalos:
In my pyramid project I just replaced my marshmallow version from 2.x to 3.0.1 and things broke when I was using **marshmallow_body_validator** by returning something like the following:

<pre>Traceback (most recent call last):
  File "/path-to-project/venv-3.6/lib/python3.6/site-packages/pyramid_debugtoolbar/toolbar.py", line 257, in toolbar_tween
    response = _handler(request)
  File "/path-to-project/venv-3.6/lib/python3.6/site-packages/pyramid_debugtoolbar/panels/performance.py", line 58, in resource_timer_handler
    result = handler(request)
  File "/path-to-project/venv-3.6/lib/python3.6/site-packages/pyramid_tm/__init__.py", line 171, in tm_tween
    reraise(*exc_info)
  File "/path-to-project/venv-3.6/lib/python3.6/site-packages/pyramid_tm/compat.py", line 36, in reraise
    raise value
  File "/path-to-project/venv-3.6/lib/python3.6/site-packages/pyramid_tm/__init__.py", line 136, in tm_tween
    response = handler(request)
  File "/path-to-project/venv-3.6/lib/python3.6/site-packages/pyramid/tweens.py", line 43, in excview_tween
    response = _error_handler(request, exc)
  File "/path-to-project/venv-3.6/lib/python3.6/site-packages/pyramid/tweens.py", line 13, in _error_handler
    response = request.invoke_exception_view(exc_info)
  File "/path-to-project/venv-3.6/lib/python3.6/site-packages/pyramid/view.py", line 767, in invoke_exception_view
    request_iface=request_iface.combined,
  File "/path-to-project/venv-3.6/lib/python3.6/site-packages/pyramid/view.py", line 667, in _call_view
    response = view_callable(context, request)
  File "/path-to-project/venv-3.6/lib/python3.6/site-packages/pyramid/viewderivers.py", line 401, in viewresult_to_response
    result = view(context, request)
  File "/path-to-project/venv-3.6/lib/python3.6/site-packages/pyramid/tweens.py", line 41, in excview_tween
    response = handler(request)
  File "/path-to-project/venv-3.6/lib/python3.6/site-packages/pyramid/router.py", line 148, in handle_request
    registry, request, context, context_iface, view_name
  File "/path-to-project/venv-3.6/lib/python3.6/site-packages/pyramid/view.py", line 667, in _call_view
[...]

PR #515, opened 2019-09-10
fixed compatibility issues of marshmallow validator with marshmallow-3.x

mamalos:
Fixes issue #514 where **validators/_marshmallow.py** brakes when using **marshmallow 3.x**.

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,7 +1,7 @@
 Sphinx
 WebTest
 colander>=1.0b1
-marshmallow>=2.0.0,<3.0.0
+marshmallow>=3.0.0rc1,<4.0
 coverage
 mock
 nose
--- a/cornice/validators/_marshmallow.py
+++ b/cornice/validators/_marshmallow.py
@@ -49,7 +49,7 @@ def _generate_marshmallow_validator(location):
         schema = _instantiate_schema(schema)
 
         class ValidatedField(marshmallow.fields.Field):
-            def _deserialize(self, value, attr, data):
+            def _deserialize(self, value, attr, data, **kwargs):
                 schema.context.setdefault('request', request)
                 deserialized = schema.load(value)
                 # marshmallow 2.x returns a tuple, 3/x will always throw
@@ -125,9 +125,15 @@ def _message_normalizer(exc, no_field_name="_schema"):
                     new_dict.update(item)
                 return {'_schema': new_dict}
         return exc.messages
-    if len(exc.field_names) == 0:
+    if hasattr(exc, 'field_names'):
+        field_names = exc.field_names
+    elif hasattr(exc, 'kwargs') and 'field_names' in exc.kwargs:
+        field_names = exc.kwargs['field_names']
+    else:
+        field_names = []
+    if len(field_names) == 0:
         return {no_field_name: exc.messages}
-    return dict((name, exc.messages) for name in exc.field_names)
+    return dict((name, exc.messages) for name in field_names)
 
 
 def validator(request, schema=None, deserializer=None, **kwargs):
--- a/tests/validationapp.py
+++ b/tests/validationapp.py
@@ -214,7 +214,7 @@ if COLANDER:
     def group_signup_post(request):
         return {'data': request.validated}
 
-    def validate_bar(node, value):
+    def validate_bar(node, value, **kwargs):
         if value != 'open':
             raise Invalid(node, "The bar is not open.")
 
@@ -368,7 +368,7 @@ if MARSHMALLOW:
         csrf_secret = marshmallow.fields.String()
 
         @marshmallow.validates_schema
-        def validate_csrf_secret(self, data):
+        def validate_csrf_secret(self, data, **kwargs):
             # simulate validation of session variables
             if self.context['request'].get_csrf() != data.get('csrf_secret'):
                 raise marshmallow.ValidationError('Wrong token')
@@ -426,7 +426,7 @@ if MARSHMALLOW:
         field = marshmallow.fields.List(marshmallow.fields.String(), many=True)
 
         @marshmallow.pre_load()
-        def normalize_field(self, data):
+        def normalize_field(self, data, **kwargs):
             if 'field' in data and not isinstance(data['field'], list):
                 data['field'] = [data['field']]
             return data
@@ -463,12 +463,12 @@ if MARSHMALLOW:
         querystring = marshmallow.fields.Nested(MRefererSchema)
 
         @marshmallow.validates_schema
-        def validate_email_length(self, data):
+        def validate_email_length(self, data, **kwargs):
             email = data['body'].get('email')
             ref = data['querystring'].get('ref')
             if email and ref and len(email) != ref:
                 raise marshmallow.ValidationError(
-                    {'email': 'Invalid email length'})
+                    {'body': {'email': 'Invalid email length'}})
 
     @m_email_service.post(
         schema=MNewsletterPayload, validators=(marshmallow_validator,))


Natim:
Don't we have a test for that? Do you know why master tests are green?

mamalos:
Ah, yeah, if you see **tests/requirements.txt**,marshmallow support was for versions <3.0.0. I guess that with this "fix" this limitation can be withdrawn.

mamalos:
Hmmm...removing `<3.0.0` from **requirements.txt** brakes various tests, so I'll dig more into it and will update my PR. 

I'll push my changes so everyone can suggest solutions about the errors.

tests/requirements.txt
```diff
@@ -1,7 +1,7 @@
 Sphinx
 WebTest
 colander>=1.0b1
-marshmallow>=2.0.0,<3.0.0
+marshmallow>=2.0.0
```
Natim:
```suggestion
marshmallow
```

mamalos:
Unfortunately things are a bit more complex than I though and a few other things are braking (as seen in the failing tests). I've corrected all of them but one, for which I opened an [issue](https://github.com/marshmallow-code/marshmallow/issues/1414) at the marshmallow project.

I hope that the guys at marshmallow will answer soon, and based on their answer I'll try resolving this issue as well.

tests/requirements.txt
sloria: Well, you probably don't want to support marshmallow 1, right? 😃 

I'd recommend

```suggestion
marshmallow>=2.0.0,4.0.0
```

Natim: If you don't specify a version it will take the latest one. However you don't know yet if 4.0 will be supported or not, so it is not the responsibility of this lib to decide.

More information about how to handle libraries dependencies in Python are available there: https://github.com/mozilla-services/servicedenuages.fr/blob/master/content/2017.02.handling-python-projects-dependencies.rst

mamalos:
Hmmm...I need some help here from somebody who knows cornice well. 

Marshmallow has changed its behaviour regarding `dict` arguments when raising **ValidationError**s, where each key is considered a field and each value is considered its error message (see [here](https://github.com/marshmallow-code/marshmallow/pull/1026)). With this in mind I am not sure how **_marshmallow.py** should change to adopt.

So, for example we see [this code](https://github.com/Cornices/cornice/blob/master/tests/validationapp.py#L471) in **validationapp.py**:
```python
        @marshmallow.validates_schema
        def validate_email_length(self, data):
            email = data['body'].get('email')
            ref = data['querystring'].get('ref')
            if email and ref and len(email) != ref:
                raise marshmallow.ValidationError(
                    {'email': 'Invalid email length'})
```

where **marshmallow.ValidationError** is passed a `dict` when called. The way marshmallow versions >= 3.0.0 behave is that the **location** is removed by the current implementation, and as a side effect, when the validation error returns to cornice there's no way of knowing where the error has occurred (ie body, schema, etc).

Since this seems to be marshmallow's wanted behaviour, then the *validator* function of **_marshmallow.py** will need to change in order to be aware about the **location** and add it appropriately when needed. 

Before doing so, however, I think that a senior's cornice dev opinion should be asked as well.

That said, even though in my current PR one marshmallow test fails, the exact same test of colander doesn't fail (see [here](https://github.com/Cornices/cornice/blob/master/tests/test_validation.py#L252)), but I think out of pure luck, because the value of `response.json['errors']` is this:
```python
[{'location': '', 'name': '', 'description': 'Invalid email length'}]
```
where **location** is empty and so is **name**; so, I think this test (and its marshmallow equivalent) should be changed to also check the **location** and **name**, unless cornice is fine with not returning either of them in this case.

So, any help/ideas would be more than welcome here!

tests/requirements.txt
leplatrem: `marshmallow>=2.0.0` seems appropriate

mamalos:
Hi guys, I need some help here. It would be best if @ergo and/or whoever else has contributed the relevant code for the colander and/or marshmallow helpers, shared their opinion (because both helpers are very similar to each other). And that's because I'm not sure that I totally follow their logic, but if I do so, then I think there's a problem somewhere.

So, if I understand correctly, there's a **_generate_marshmallow_validator** function (and a **_generate_colander_validator** for colander, respectively) where the supplied validation schema is "replaced" (for the sake of brevity) with a new schema that contains one single **ValidatedField** whose deserialisation function effectively deserialises the input based on the user supplied schema.

When reading the code, I expect that if a **ValidationError** is raised from the user-supplied schema, this should be done through the `deserialized = schema.load(value)` clause of **ValidatedField**'s[ _deserialize function](https://github.com/Cornices/cornice/blob/master/cornice/validators/_marshmallow.py#L54). But when debugging the tests, I've met cases (like the last one that I'm mentioning) where this was not the case, and **ValidationErrors** were raised from **validator**'s [function similar clause](https://github.com/Cornices/cornice/blob/master/cornice/validators/_marshmallow.py#L166) without reaching the above-mentioned **ValidatedField**'s code first.

So, even though I can move on and correct the tests to produce the expected outcomes given the [response](https://github.com/marshmallow-code/marshmallow/issues/1414) I was given from the marshmallow team, I am not 100% sure that everything works as expected, at least from the marshmallow helper module.

Hence, again, any help is more than welcome here.

mamalos:
Hi again,

I'll try to write it as simple as I can, and this needs feedback from someone who understands cornice's **validators** functionality more than me:

Should [ValidatedField._deserialize()](https://github.com/Cornices/cornice/blob/master/cornice/validators/_marshmallow.py#L54) be called whenever **validator()**'s [schema.load(cstruct)](https://github.com/Cornices/cornice/blob/master/cornice/validators/_marshmallow.py#L166) is called or not?

My understanding is that is should, but this is not the case (and not only for the test that brakes). And if it should, then more stuff will have to be re-written in **validators/_marshmallow.py** because things are breaking inside marshmallow and not inside cornice (and not because of marshmallow...).

So, @natim and/or @leplatrem (and/or whoever can help), could you please ping someone who can answer this? Because without this question being answered (and some insight be given so as to either help/direct me overcome the problem, or someone correcting it themselves) marshmallow support in cornice is unfortunately buggy (marshmallow tests fail on versions >=3, and don't fail -but the tests themselves are not very to-the-point- on versions <3).

Btw, the tests that are failing are a bit non-realistic or better say they don't test validation the way it will be used in cornice. And that's because they declare a schema that expects two fields, **body** and **querystring**, and then validation is performed with ```marshmallow.validates_schema``` decorator, whereas in most cases, developers using cornice would create schemas with fields being those that their API expects to serialise/deserialise and mainly use ```marshmallow.validates``` decorator instead (marshmallow.validates_schema could also be used, but again, in the context that it's designed to be used, so for the API's fields).

mamalos:
Hey guys (@Natim, @leplatrem) I need some help here. Marshmallow 2 and 3 are in some places mutually exclusive in terms of the provided output and behaviour, so there's no way of achieving 100% coverage when only one version of marshmallow is used in the tests. Moreover, due to the same reasons, the error message I corrected in order for tests to pass when using marshmallow 3.x, it obviously fails the tests under marshmallow 2.x.

I suppose that people would want to use both marshmallow versions with cornice, but in that case somebody should tell me how to address the coverage issue. Is there a way to run the tests for two versions of marshmallow module and use the union of their coverage? Or is there a better way to resolve this issue (which I suppose exists)?

Natim:
You should be able to create a tox matrix with both and use @skip to skip one tests when using marshmallow  2 and the others when using marshmallow 3. However we could also decide to drop support for v2 and keep only support for v3

mamalos:
Thanks for the feedback @Natim, but let me just rephrase in order to see if I've understood you correctly -because I'm new to tox:

I'll use tox matrix to create all the necessary tests with the various required/supported versions and in each such test the non-relevant sections will have to be removed using @Skip, right?

mamalos:
Hey, @Natim, I'd like to write the tests and finally close this PR but I'm not sure how to do it exactly. You mentioned @skip in your previous answer, but github translates it to the user Skip, so I'm not sure which `@skip/@Skip` you're talking about (nosetest's, unittest's, etc.). Could you please clear this out?

And also, could you please explain how to do the conditional tests (or provide a link to some example)? Ie check the `TOX_ENV_NAME` environmental variable to see which tox environment the test is using and skip it based on its value?

Thanks.

mamalos:
Hmm...OK @Natim, I think I found how to do the tests. I'll check the `TOX_ENV_NAME` env variable in **test_validation.py** and create a new `@skip_if_marshmallow_2` and `@skip_if_marshmallow_3` decorators (based on how `skip_if_no_colander` and `skip_if_no_marshmallow` are created) depending on its value.

leplatrem:
Can we drop support for marshmallow 2 and proceed with this?

mamalos:
Hi @leplatrem. Dropping support for marshmallow v2 would definitely help in order for the tests to pass, but it wouldn't solve the issues I'm mentioning earlier in this post (on 1 Oct 2019 and 21 Oct 2019). And the things I'm mentioning imply that the behaviour of marshmallow within cornice -at least to my understanding when I was writing my comments, because now I don't remember a thing- is not the one that should have been expected.

So, I would really appreciate if someone from the people I'm mentioning in my posts would also take a look into it, but even if not, I'll try to find a day or two to fix this thing because it's a pity to leave it in this state.

Issue #511, opened 2019-07-19
[Feature request] It would be nice if one could pass marshmallow's keyword arguments when creating a schema object.


mamalos:
Hi all,

When using the **marshmallow_body_validator** (or its siblings), one has to provide a class object (of a schema) rather than an instance. That's understandable, but it limits its usage within views.

For example, if someone wants to use **many=True**, as @ergo suggested in IRC, one could subclass the original class and pass **many=True** through its constructor, hence something like this:

```python
class MyCustomManySchema(MySchema):
    def __init__(self, *args, **kwargs):
        kwargs['many'] = True
        Schema.__init__(self, *args, **kwargs)
```

which is a cool workaround, but adds more classes, boilerplate, complexity, etc.

Moreover, all other marsmallow nifty features that are added through it's constructor (like **only**, **partial**, etc) have to be treated accordingly (if applicable).

I tried changing **_marshmallow.py** **validator** to add one more _dict_ argument that could have been used to store the keyword arguments when instantiating a schema, but it didn't work (I think it was messing up the meta class defined in **_validator()**). Here's a patch from what I tried that **DIDN'T** work:
```diff
--- _marshmallow.py	2019-07-19 14:54:15.308854413 +0300
+++ _marshmallow.py.test	2019-07-19 14:46:29.880363619 +0300
@@ -19,7 +19,8 @@
     :rtype: callable
     """
 
-    def _validator(request, schema=None, deserializer=None, **kwargs):
+    def _validator(request, schema=None, deserializer=None, schema_kwargs={},
+                   **kwargs):
         """
         Validate the location against the schema defined on the service.
 
@@ -46,7 +47,7 @@
         if schema is None:
             return
 
-        schema = _instantiate_schema(schema)
+        schema = _instantiate_schema(schema, schema_kwargs)
 
         class ValidatedField(marshmallow.fields.Field):
             def _deserialize(self, value, attr, data):
@@ -96,7 +97,8 @@
             """A schema to validate the request's location attributes."""
             pass
 
-        validator(request, RequestSchema, deserializer, **kwargs)
+        validator(request, RequestSchema, deserializer, schema_kwargs,
+                  **kwargs)
         request.validated = request.validated.get(location, {})
 
     return _validator
@@ -130,7 +132,8 @@
     return dict((name, exc.messages) for name in exc.field_names)
 
 
-def validator(request, schema=None, deserializer=None, **kwargs):
[...]

Issue #501, opened 2018-12-21
Passing kwargs to Schema


Philipp-g:
Since https://github.com/Cornices/cornice/pull/482 Cornice only accepts a Marshmallow Schema instead of a Schema instance. So I was wondering if it is possible to pass arguments to the schema instantiation.

I just glanced through the code and this suggests that it is not possible right now.
https://github.com/Cornices/cornice/blob/0359fddc01915756f77acecf4d6c5f6297940f29/cornice/validators/_marshmallow.py#L188-L192

leplatrem:
ping @ergo :)

PR #516, opened 2019-09-12
Support kwargs in marshmallow validator

mamalos:
With this patch, keyword arguments are supported when instantiating the marshmallow schema (#511).

This is an example of how to use it:

```python
def my_validator(request, **kwargs):
    if request.method == 'POST':
        sch = MyPOSTSchema
        kwargs['schema_kwargs'] = {'many': True}
    elif request.method == 'PATCH':
        sch = MyPATCHSchema
    kwargs['schema'] = sch
    return marshmallow_body_validator(request, **kwargs)

class MyView(object):
    @view(validators=(my_validator,))
    def patch(self):
        return self._update()
```

So, if someone passes **schema_kwargs** dictionary in a marshmallow validator, it is expanded in the schema's constructor when it is instantiated.

Pinging @ergo for a second opinion.

### DIFF
--- a/cornice/validators/_marshmallow.py
+++ b/cornice/validators/_marshmallow.py
@@ -52,7 +52,7 @@ def _generate_marshmallow_validator(location):
         # see if the user wants to set any keyword arguments for their schema
         schema_kwargs = {}
         if 'schema_kwargs' in kwargs:
-            schema_kwargs = kwargs['schema_kwargs']
+            schema_kwargs = kwargs.get('schema_kwargs', {})
         schema = _instantiate_schema(schema, **schema_kwargs)
 
         class ValidatedField(marshmallow.fields.Field):
--- a/cornice/validators/_marshmallow.py
+++ b/cornice/validators/_marshmallow.py
@@ -50,9 +50,7 @@ def _generate_marshmallow_validator(location):
             return
 
         # see if the user wants to set any keyword arguments for their schema
-        schema_kwargs = {}
-        if 'schema_kwargs' in kwargs:
-            schema_kwargs = kwargs.get('schema_kwargs', {})
+        schema_kwargs = kwargs.get('schema_kwargs', {})
         schema = _instantiate_schema(schema, **schema_kwargs)
 
         class ValidatedField(marshmallow.fields.Field):
--- a/cornice/validators/_marshmallow.py
+++ b/cornice/validators/_marshmallow.py
@@ -27,7 +27,7 @@ def _generate_marshmallow_validator(location):
         the ``request.validated`` attribute.
 
         Keyword arguments to be included when initialising the marshmallow
-        schema can be passed as a dict in kwargs['schema_kwargs'] variable.
+        schema can be passed as a dict in ``kwargs['schema_kwargs']`` variable.
 
         .. note::
 
--- a/cornice/validators/_marshmallow.py
+++ b/cornice/validators/_marshmallow.py
@@ -26,6 +26,9 @@ def _generate_marshmallow_validator(location):
         The content of the location is deserialized, validated and stored in
         the ``request.validated`` attribute.
 
+        Keyword arguments to be included when initialising the marshmallow
+        schema can be passed as a dict in kwargs['schema_kwargs'] variable.
+
         .. note::
 
             If no schema is defined, this validator does nothing.
@@ -46,10 +49,14 @@ def _generate_marshmallow_validator(location):
         if schema is None:
             return
 
-        schema = _instantiate_schema(schema)
+        # see if the user wants to set any keyword arguments for their schema
+        schema_kwargs = {}
+        if 'schema_kwargs' in kwargs:
+            schema_kwargs = kwargs['schema_kwargs']
+        schema = _instantiate_schema(schema, **schema_kwargs)
 
         class ValidatedField(marshmallow.fields.Field):
-            def _deserialize(self, value, attr, data):
+            def _deserialize(self, value, attr, data, **kwargs):
                 schema.context.setdefault('request', request)
                 deserialized = schema.load(value)
                 # marshmallow 2.x returns a tuple, 3/x will always throw
@@ -185,8 +192,17 @@ def validator(request, schema=None, deserializer=None, **kwargs):
         request.validated.update(deserialized)
 
 
-def _instantiate_schema(schema):
+def _instantiate_schema(schema, **kwargs):
+    """
+    Returns an object of the given marshmallow schema.
+
+    :param schema: The marshmallow schema class with which the request should
+        be validated
+    :param kwargs: The keyword arguments that will be provided to the
+        marshmallow schema's constructor
+    :return: The object of the marshmallow schema
+    """
     if not inspect.isclass(schema):
         raise ValueError('You need to pass Marshmallow class instead '
                          'of schema instance')
-    return schema()
+    return schema(**kwargs)
--- a/tests/validationapp.py
+++ b/tests/validationapp.py
@@ -348,16 +348,6 @@ if MARSHMALLOW:
             unknown = EXCLUDE
         username = marshmallow.fields.String()
 
-    class MSignupGroupSchema(marshmallow.Schema):
-        class Meta:
-            strict = True
-            unknown = EXCLUDE
-        username = marshmallow.fields.String()
-
-        def __init__(self, *args, **kwargs):
-            kwargs['many'] = True
-            marshmallow.Schema.__init__(self, *args, **kwargs)
-
     import random
 
     class MNeedsContextSchema(marshmallow.Schema):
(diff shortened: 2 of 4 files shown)


mamalos:
Hi @Natim, no it's not a duplicate; #515 fixes a compatibility issue with marshmallow v.3.x (which prohibited its use through cornice's marshmallow helper), this one tries to add some functionality to marshmallow validator, so that one can exploit many more of its features (as I just saw it was also requested as a feature in #501 around 7 months ago).

I'll have to see what broke things in the tests and see if I can fix it.

Natim:
Ok thank you for the clarification, I have to say I don't know much about marshmallow support in Cornice.

mamalos:
Np, me neither :). 

In terms of failing tests, as I saw, I need to add some tests to keep coverage at 100%. Will do it later when I find some free time.

mamalos:
OK, I've added some comments in **validators/_marshmallow.py** of how it now supports **schema_kwargs** variable from views.

I've replaced one of the existing tests that used an extra schema that added **many=True** in its constructor, with one that uses the new functionality (ie passing ```{'many': True}``` as a **schema_kwargs** variable when calling **marshmallow_body_validator**) so that the tests meet the 100% coverage requirement.

mamalos:
If this PR gets accepted, we should probably change cornice documentation of this functionality. Not sure if the docs (https://cornice.readthedocs.io) are maintained this repo, if so I have no problem of updating the relevant section.

Natim:
Yes the docs are there: https://github.com/Cornices/cornice/tree/master/docs

tests/validationapp.py
```diff
@@ -383,8 +373,15 @@ def m_bound_post(request):
...
-        schema=MSignupGroupSchema, validators=(marshmallow_body_validator,))
+    # callback that returns a validator with keyword arguments for marshmallow
+    # schema initialisation. In our case it passes many=True to the desired
+    # schema
+    def get_my_marshmallow_validator_with_kwargs(request, **kwargs):
+        kwargs['schema'] = MSignupSchema
+        kwargs['schema_kwargs'] = {'many': True}
+        return marshmallow_body_validator(request, **kwargs)
+
+    @m_group_signup.post(validators=(get_my_marshmallow_validator_with_kwargs,))
```
Natim: What about: ?

```suggestion
        @m_group_signup.post(
        schema=MSignupGroupSchema, schema_kwargs={"many": True}, validators=(marshmallow_body_validator,))
```

mamalos: Ah, OK, I didn't know that the arguments passed in a service's **post()** call (or other analogous calls) are also passed to the declared validators. If that's the case then your suggested syntax is more straight forward than the one I suggested, so I'm cool with that. 

My only comment is that in order to test the new code correctly, one should use `MSignupSchema` instead of `MSignupGroupSchema`, because the latter class has `many=True` in its constructor. The whole idea behind my suggestion of passing `schema_args`is exactly about removing the need of having separate schema marshmallow classes when one just needs one schema class and tailors it by exploiting marshmallow's ability of changing many of the original schema's properties just by passing the appropriate arguments in the schema's object initialisation.

leplatrem:
What's the status of this?

Also, could you please add a CHANGELOG entry and yourself to the CONTRIBUTORS?

mamalos:
Hi, I need to check if I've included the changes suggested by @Natim, push them in case I haven't done so, and I think we're done with it.

And as you suggested, I should update the CHANGELOG and the CONTRIBUTORS files accordingly.

mamalos:
OK, I've updated `CONTRIBUTORS.txt` and `CHANGES.txt` and also tried to see if @Natim suggestion would work but unfortunately it didn't.

@Natim, when trying to pass your version of the code, many tests are failing. Here's the error's I'm getting in one of such failing test:
```Error
  File "/usr/lib/python2.7/unittest/case.py", line 329, in run
    testMethod()
  File "/home/mamalos/my_programs/contrib/github/cornice/tests/test_validation.py", line 497, in test_bound_schema_multiple_calls
    app = self.make_ordinary_app()
  File "/home/mamalos/my_programs/contrib/github/cornice/tests/test_validation.py", line 477, in make_ordinary_app
    return TestApp(main({}))
  File "/home/mamalos/my_programs/contrib/github/cornice/tests/validationapp.py", line 524, in main
    return CatchErrors(config.make_wsgi_app())
  File "/home/mamalos/my_programs/contrib/github/cornice/.tox/py27/local/lib/python2.7/site-packages/pyramid/config/__init__.py", line 891, in make_wsgi_app
    self.commit()
  File "/home/mamalos/my_programs/contrib/github/cornice/.tox/py27/local/lib/python2.7/site-packages/pyramid/config/actions.py", line 152, in commit
    self.action_state.execute_actions(introspector=self.introspector)
  File "/home/mamalos/my_programs/contrib/github/cornice/.tox/py27/local/lib/python2.7/site-packages/pyramid/config/actions.py", line 293, in execute_actions
    action = next(action_iter, None)
  File "/home/mamalos/my_programs/contrib/github/cornice/.tox/py27/local/lib/python2.7/site-packages/pyramid/config/actions.py", line 425, in resolveConflicts
    discriminator = undefer(action['discriminator'])
  File "/home/mamalos/my_programs/contrib/github/cornice/.tox/py27/local/lib/python2.7/site-packages/pyramid/registry.py", line 301, in undefer
    v = v.resolve()
  File "/home/mamalos/my_programs/contrib/github/cornice/.tox/py27/local/lib/python2.7/site-packages/pyramid/registry.py", line 292, in resolve
    return self.value
  File "/home/mamalos/my_programs/contrib/github/cornice/.tox/py27/local/lib/python2.7/site-packages/pyramid/decorator.py", line 43, in __get__
    val = self.wrapped(inst)
  File "/home/mamalos/my_programs/contrib/github/cornice/.tox/py27/local/lib/python2.7/site-packages/pyramid/registry.py", line 287, in value
    result = self.func()
  File "/home/mamalos/my_programs/contrib/github/cornice/.tox/py27/local/lib/python2.7/site-packages/pyramid/config/views.py", line 940, in discrim_func
    self._check_view_options(**dvals)
  File "/home/mamalos/my_programs/contrib/github/cornice/.tox/py27/local/lib/python2.7/site-packages/pyramid/config/views.py", line 1209, in _check_view_options
    raise ConfigurationError('Unknown view options: %s' % (kw,))
ConfigurationError: Unknown view options: {'schema_kwargs': {'many': True}}
```

Here's the code I've tried:
```python
    @m_group_signup.post(
        schema=MSignupSchema, schema_kwargs={'many': True},
        validators=(marshmallow_body_validator,))
    def m_group_signup_post(request):
        return {'data': request.validated}
```

Not sure why it happens -your suggestion seems very reasonable to me too-, but unless you want to dig it any futher, I suggest we keep the current version that's passing the tests - if you're cool with that obviously.

mamalos:
Hi there (@natim, @leplatrem),

Maybe it's time for somebody to review and accept the change?

cornice/validators/_marshmallow.py
```diff
@@ -46,10 +49,14 @@ def _validator(request, schema=None, deserializer=None, **kwargs):
         if schema is None:
             return
 
-        schema = _instantiate_schema(schema)
+        # see if the user wants to set any keyword arguments for their schema
+        schema_kwargs = {}
+        if 'schema_kwargs' in kwargs:
+            schema_kwargs = kwargs.get('schema_kwargs', {})
```
leplatrem: Sorry this code suggestion thing didn't work as I expected. You can remove the above lines

mamalos:
Hi @leplatrem and thanks for the comments and suggestions.

In terms of class vs instance, I haven't changed anything in the current implementation. Marshmallow is different from colander in this matter. It allows you to reuse the same validator class and tailor it appropriately upon instantiation (using constructor's args/kwargs), which is one of its very strengths. 

So, despite the fact that the provided PR is not very "elegant" (having a `schema_kwargs` as a keyword argument is not the "pretiest" thing), but it allows for proper integration of marshmallow with cornice and helps very much in code reuse.

Once this PR (and #515) are complete and accepted, I'll also update the documentation in order for people to know they can use it.

cornice/validators/_marshmallow.py
mamalos: Sorry, you lost me in this. The truth is that I accepted this change without seeing it too deeply, I thought it was in a different spot. Now I'm a bit confused, has it been merged with my code or not?

leplatrem: Just that these 3 lines can be replaced by the last one

`schema_kwargs = kwargs.get('schema_kwargs', {})`

mamalos: Is it OK now? I'm not familiar with these github's features (merging suggested changes and editing files within github), that's why I'm asking. 

So, I've edited the file within github to only contain `schema_kwargs = kwargs.get('schema_kwargs', {})` and I suppose it's OK now, but I'm not sure I saw the checks to have run again.

PR #518, opened 2019-10-16
Frame was always set to 2 for resource decorator.

dugagjin:
Which frame should be looked in defaults to 2 but was always 2 for resource decorator. This commit fixes this.

### DIFF
--- a/cornice/resource.py
+++ b/cornice/resource.py
@@ -115,7 +115,7 @@ def add_resource(klass, depth=1, **kw):
                         klass.__name__.lower())
         service_name = prefix + service_name
         service = services[service_name] = Service(name=service_name,
-                                                   depth=2, **service_args)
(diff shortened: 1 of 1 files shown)


Natim:
I remember this caught my attention in the past, not sure why we set it to 2.

Do you have a case in mind showing the issue? I'd like to add a test for this specific case.

cornice/resource.py
```diff
@@ -115,7 +115,7 @@ class User(object):
                         klass.__name__.lower())
         service_name = prefix + service_name
         service = services[service_name] = Service(name=service_name,
-                                                   depth=2, **service_args)
+                                                   depth=depth, **service_args)
```
leplatrem: Nice catch! But could you please restore the default to 2 in the signature?

leplatrem:
Merged in #556

PR #520, opened 2019-10-23
Drop Python 2 support

leplatrem:

PR #531, opened 2020-04-01
Drops simplejson as requirement

devilicecream:
Changes the name of the default renderer for Services to `'cornicejson'`. For retrocompatibility,
when simplejson is installed, the default renderer stays as `'simplejson'` and behaves
the same way it used to. Otherwise, the default pyramid serializer (`json.dumps(...)`) is used.

### DIFF
--- a/cornice/service.py
+++ b/cornice/service.py
@@ -12,8 +12,13 @@ from cornice.validators import (
 )
 import venusian
 
-from cornice.util import is_string, to_list, json_error_handler, func_name
-
+from cornice.util import (
+    is_string,
+    to_list,
+    json_error_handler,
+    func_name,
+    default_bytes_adapter,
+)
 
 SERVICES = []
 
@@ -150,7 +155,7 @@ class Service(object):
     :meth:`~put`, :meth:`~options` and :meth:`~delete` are decorators that can
     be used to decorate views.
     """
-    renderer = 'simplejson'
+    renderer = 'cornicejson'
     default_validators = DEFAULT_VALIDATORS
     default_filters = DEFAULT_FILTERS
 
@@ -255,7 +260,8 @@ class Service(object):
             default_error_handler = json_error_handler(simplejson.dumps,
                                                        {"use_decimal": True})
         else:
-            default_error_handler = json_error_handler(json.dumps)
+            default_error_handler = json_error_handler(
+                json.dumps, {"default": default_bytes_adapter})
         # Allow custom error handler
         arguments['error_handler'] = conf.pop(
             'error_handler',
--- a/cornice/service.py
+++ b/cornice/service.py
@@ -2,6 +2,8 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 import functools
+import json
+
 from pyramid.exceptions import ConfigurationError
 from pyramid.response import Response
 from cornice.validators import (
@@ -10,7 +12,7 @@ from cornice.validators import (
 )
 import venusian
 
-from cornice.util import is_string, to_list, json_error, func_name
+from cornice.util import is_string, to_list, json_error_handler, func_name
 
 
 SERVICES = []
@@ -155,6 +157,10 @@ class Service(object):
     mandatory_arguments = ('renderer',)
     list_arguments = ('validators', 'filters', 'cors_headers', 'cors_origins')
 
+    @classmethod
+    def configure_default_renderer(cls, renderer="simplejson"):
+        cls.renderer = renderer
+
     def __repr__(self):
         return u'<Service %s at %s>' % (
             self.name, self.pyramid_route or self.path)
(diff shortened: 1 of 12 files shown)


devilicecream:
Closes #418

leplatrem:
Thanks for the efforts on this!

I feel a bit uncomfortable with the amount of changes you had to do. Do you personally need backward compatibility? It looks like the changes are scattered all over and to be honest I don't think it will help us as maintainers :| 

I'd like to ship #520 therefore I wouldn't mind introducing some breaking changes here. Especially if it simplifies the code base! 
(If maintainers are motivated to maintain a 2.7 branch, they are more than welcomed :))

What are your guts feelings?

devilicecream:
> Thanks for the efforts on this!
> 
> I feel a bit uncomfortable with the amount of changes you had to do. Do you personally need backward compatibility? It looks like the changes are scattered all over and to be honest I don't think it will help us as maintainers :|
> 
> I'd like to ship #520 therefore I wouldn't mind introducing some breaking changes here. Especially if it simplifies the code base!
> (If maintainers are motivated to maintain a 2.7 branch, they are more than welcomed :))
> 
> What are your guts feelings?

hey @leplatrem! thanks for taking a look to it. I agree there's a lot of changes needed, but the retrocompatibility is not just for python2.7 support, but mostly to keep the previous behavior without breaking current implementations using simplejson.
I feel like, as a user of Cornice, I wouldn't like an update that would break my code without manually overriding the default json renderer, that's why I went this way.
Additionally, since at my company we're using an older version of cornice, I would like the patch to be applicable to it without breaking changes, but that's another story of course.
Let's wait for the mantainers feedback anyway!

cornice/service.py
```diff
@@ -241,10 +252,21 @@ def get_arguments(self, conf=None):
...
+        if self.renderer == "simplejson":
+            # XXX: keep retrocompatibility
+            # when using 'simplejson' as renderer (default)
+            import simplejson
+
+            default_error_handler = json_error_handler(simplejson.dumps,
+                                                       {"use_decimal": True})
+        else:
+            default_error_handler = json_error_handler(
+                json.dumps, {"default": default_bytes_adapter})
```
leplatrem, line 264: Could this be done in `.configure_default_renderer()` ?

cornice/service.py
```diff
@@ -148,13 +155,17 @@ class Service(object):
...
-    renderer = 'simplejson'
+    renderer = 'cornicejson'
     default_validators = DEFAULT_VALIDATORS
     default_filters = DEFAULT_FILTERS
 
     mandatory_arguments = ('renderer',)
     list_arguments = ('validators', 'filters', 'cors_headers', 'cors_origins')
 
+    @classmethod
+    def configure_default_renderer(cls, renderer="simplejson"):
```
leplatrem, line 166: no need for default value for `renderer` parameter

cornice/util.py
```diff
@@ -84,20 +96,35 @@ def to_list(obj):
...
         body = {'status': 'error', 'errors': errors}
-        Response.__init__(self, simplejson.dumps(body, use_decimal=True))
+        Response.__init__(self, renderer(body, **renderer_args))
         self.status = status
         self.content_type = 'application/json'
 
 
-def json_error(request):
-    """Returns an HTTPError with the given status and message.
+def json_error_handler(renderer, renderer_args=None):
```
leplatrem, line 114: should maybe be called `json_error_handler_factory` ?

tests/support.py
```diff
@@ -3,21 +3,33 @@
...
 from pyramid import testing
 
 from cornice.errors import Errors
 
+try:
+    import simplejson
+
+    SIMPLEJSON = True
+except ImportError:
+    SIMPLEJSON = False
```
leplatrem, line 26: this could be done only once in `utils.py`, `HAS_SIMPLEJSON = ...`

leplatrem:
> I feel like, as a user of Cornice, I wouldn't like an update that would break my code without manually overriding the default json renderer, that's why I went this way.

Yes, manual upgrades are painful and I understand your position. I find it acceptable in that case, since we would bump a major, with several breaking changes (drop of python 2.7). An upgrade section in the docs would give you the instructions for your project (like these https://cornice.readthedocs.io/en/latest/upgrading.html). Upgrading a library to a major version, usually implies reading the changelog.

We are not many maintainers, and all with limited time. Maintaining OSS projects takes time and efforts, and as you can see we don't respond and fix every coming issue or suggestion in the minute.
That's partly why, generally speaking, I'd like to strip down Cornice and leave as little code and hacks as possible. So if droping simplejson supports means removing some code I'd happilly merge it. But if dropping simplejson means adding more code, branches, patching etc. then I will be more reluctant :) 
Does it make sense?

cornice/service.py
devilicecream: I agree. To be fair, I had a hard time understanding why is there a `renderer` object attached to the Service class, but then the same renderer is not for handling errors as well. 

Would it actually be better to find a way of using the `Service.renderer` also to perform the rendering of errors?

cornice/service.py
devilicecream: right. comes from an earlier version of the patch

tests/support.py
devilicecream: that's right, I'll fix this

cornice/util.py
devilicecream: Definitely, I'll update this once we figure out what to do with error handlers, as for the comment above.

devilicecream:
Given your feedback on the PR, and given the current status of the renderer implementation, I have another resolution strategy, that looks more as an additional feature that a simple bugfix though.

What if:
1. we make an extendable `CorniceRenderer` class that also works as a pyramid renderer (very similar to the current implementation of `_JsonRenderer`, but extendable by the cornice user) 
2. make that implementation use the standard `json.dumps`
3. have the error serialization handled by the same renderer too
4. (optional) add an implementation using `simplejson` that could be used automatically if `simplejson` is installed

leplatrem:
Thanks! I like your idea! 

It would help us have everything related to serializers centralized in one place :+1: 

Do you prefer us to wait before merging #520?

devilicecream:
> Do you prefer us to wait before merging #520

That's really up to you! The only conflict I see is in the `setup.py`, cause we're both removing a dependency, but I'm happy to rebase after you merge #520 .

I'm finishing up the tests in #533 , if you want to take a look that's the solution we discussed above. Looks much better than this one for sure!

leplatrem:
Dropped in favor of #533

PR #532, opened 2020-04-02
Replace nose by pytest

leplatrem:


leplatrem:
@delijati Do you have an idea why your test about imperative resource would fail when switching from Nose to PyTest?

delijati:
Hi @leplatrem it seams the test isolation somehow does not work properly under pytest. I deactivated all other tests and run just the one failing and it worked :/

delijati:
Btw. i have no clue what this is testing `Test that we don't fail Pyramid's conflict detection when using a manually` It is just too long ago that i wrote that :/

delijati:
made my day `@pytest.mark.skip(reason="This test fails when ran with pytest, and it's too mysterious for now")` 

sorry i still have no idea why it did not work in pytest

PR #533, opened 2020-04-16
Refactors the rendering machinery of Cornice.

devilicecream:
- Makes the default cornice renderer a subclass of :class:`pyramid.renderers.JSON`, so the default rendering machinery of pyramid is still supported, while keeping the Content-Type handling logic of cornice.
- Exposes the rendering logic in :method:`cornice.renderers.CorniceRenderer.render()`, so that it's customizable if needed.
- Rationalizes the error rendering on :class:`cornice.service.Service`, so that it uses the registered render to render `request.errors`, instead of a custom function.

**BREAKING**: the default renderer does not use `simplejson.dumps` by default anymore, so the requirement has been dropped.

Mentions #531

### DIFF
--- a/cornice/__init__.py
+++ b/cornice/__init__.py
@@ -5,8 +5,8 @@ import logging
 import pkg_resources
 from functools import partial
 
-from cornice import util
 from cornice.errors import Errors  # NOQA
+from cornice.renderer import CorniceRenderer
 from cornice.service import Service   # NOQA
 from cornice.pyramidhook import (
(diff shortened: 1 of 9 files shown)


devilicecream:
Closes #418

devilicecream:
@leplatrem whenever you're free to review this. thanks! 👍

devilicecream:
@leplatrem whenever you're free to review this. thanks! 👍

PR #537, opened 2020-05-25
Set python_requires>=3.5

Nicusor97:
It will prevent users with Python 2.7 from downloading a Sdist version that cannot be built.

Note: supporting python_requires requires setuptools>=24.2.0 and pip>=9.0.0 to benefit from it.
Details here: https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires


leplatrem:
Do you think I should release a dot version with this change?

Nicusor97:
@leplatrem  From my point of view I think will be useful to have a dot version with this change.  On my side, I added a pin until we will have a new `cornice` version with python_requires='>=3.5'

leplatrem:
Done! Thanks!

PR #543, opened 2020-08-12
Rename default renderer to cornicejson

leplatrem:
Some extensions like cornice.ext.swagger assume that JSON renderers
would contain the word ``"json"``.

Renaming this default renderer does not harm and kind of makes sense.

It does not mean we shouldn't fix the extensions to remove these sorts
of assumptions and be more robust.


leplatrem:
> Breaking change?

It was never documented, so I would be tempted to say «no» (?)

PR #544, opened 2020-08-23
Copy wrapped resource's attributes into the service

spack971:
When a Service is generated from a Resource, it losses all the attributes from the resource (specifically the docstring). This change ensures that the wrapping service looks like it wrapped resource.

This is particularly useful since when I list the services registered by Cornice (`config.registry.cornice_services`), I need to have access to the original docstring.

Since the service is based on the resource, it makes also sense to mimic its attributes.

### DIFF
--- a/tests/test_resource.py
+++ b/tests/test_resource.py
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
 # You can obtain one at http://mozilla.org/MPL/2.0/.
 import json
+import functools
 from unittest import mock
 
 from pyramid import testing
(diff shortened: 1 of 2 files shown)


leplatrem (commented):
Thanks!

Could you please add a test?

spack971:
Added a test case, let me know if that works for you.

tests/test_resource.py
```diff
@@ -219,6 +222,23 @@ def test_acl_support_authenticated_allowed_thing_get(self):
...
+        resources = {
+            'thing_service': Thing,
+            'user_service': User,
+        }
+
+        for name, service in (
+            (x.name, x) for x in self.config.registry.cornice_services.values()
+            if x.name in resources
+        ):
+            for attr in functools.WRAPPER_ASSIGNMENTS:
```
leplatrem, line 235: TIL `WRAPPER_ASSIGNMENTS` :) :+1:

leplatrem (approved):

PR #545, opened 2020-08-26
Fix Colander validator missing body, path, etc.

antoineleclair:
The issue was introduced with commit
40ed37a2f453269ab85a4376914e893cf05d89c8

Validators that had attributes like `body` was broken
(`body` was missing in `request.validated`). For example:

```python
class EmailContentSchema(MappingSchema):
    subject = SchemaNode(String())
    body = SchemaNode(String())

@send_email_service.post(
    schema=EmailContentSchema(),
    validators=(colander_body_validator,),
)
def send_email_post(request):
    send_email(
        request.validated["subject"],
        request.validated["body"] # was missing
    )
    return HTTPOk()
```

The issue was with `body`, `headers`, `path`, `querystring` when using:
* `body_validator`
* `headers_validator`
* `path_validator`
* `querystring_validator`

### DIFF
--- a/cornice/validators/_colander.py
+++ b/cornice/validators/_colander.py
@@ -74,7 +74,8 @@ def _generate_colander_validator(location):
         validator(request, RequestSchema(), deserializer, **kwargs)
         validated_location = request.validated.get(location, {})
         request.validated.update(validated_location)
-        request.validated.pop(location, None)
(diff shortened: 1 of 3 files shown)


leplatrem (approved):
Good catch! Thanks!

PR #553, opened 2021-03-09
Remove references to pyramid.compat

fizyk:
That module was deprecated since pyramid 1.7, and got removed in pyramid 2.
    It was a compatibility layer for python 2 which cornice does not support since version 5


leplatrem (approved):
:+1: Thanks!

PR #558, opened 2021-04-28
Fix apply_filters where service defined by pyramid_route

dionpeters:
Use current_service to determine cornice service in pyramidhook apply_filter, previously cornice service was not found if added via pyramid_route definition.

### DIFF
--- a/CHANGES.txt
+++ b/CHANGES.txt
@@ -9,6 +9,10 @@ CHANGELOG
 
 - Drop support of Marshmallow < 3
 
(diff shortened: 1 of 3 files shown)


leplatrem (approved):

leplatrem:
Thanks! Good catch!

leplatrem:
Could you please update the CHANGELOG and add yourself to CONTRIBUTORS? Thanks!

leplatrem (approved):

PR #561, opened 2021-07-08
fix: eliminate deprecated marshmallow syntax

asthasr:
After setting up a skeleton app, I was receiving a warning while running tests:

```
RemovedInMarshmallow4Warning: Passing field metadata as a keyword arg is deprecated. Use the explicit `metadata=...` argument instead.
```
The change here seems to fix the issue.


leplatrem (approved):
Thank you!

PR #576, opened 2023-07-13
Update test matrix for the new generation of Python versions.

Natim:


Natim:
Back to green

image


Something is strange regarding the coverage config though.

PR #606, opened 2026-03-28
Support marshmallow 3 and 4

tomascorrea:
> 🚀 **Maintainer Wanted!** This project is looking for a new steward to keep it alive! 🌱 If you still care for this project 🛠️, drop a comment or a PR removing this warning! 🎉✨

### DIFF
--- a/src/cornice/validators/_marshmallow.py
+++ b/src/cornice/validators/_marshmallow.py
@@ -83,9 +83,7 @@ def _generate_marshmallow_validator(location):
                 :type dct: dict
                 """
 
-                class_attrs[location] = ValidatedField(
-                    required=True, data_key=location
-                )
+                class_attrs[location] = ValidatedField(required=True, data_key=location)
                 class_attrs["Meta"] = Meta
                 return type(name, bases, class_attrs)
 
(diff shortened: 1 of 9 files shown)


src/cornice/validators/_marshmallow.py
```diff
@@ -51,13 +56,12 @@ def _validator(request, schema=None, deserializer=None, **kwargs):
 
         class ValidatedField(marshmallow.fields.Field):
             def _deserialize(self, value, attr, data, **kwargs):
-                schema.context.setdefault("request", request)
+                cornice_request.set(request)
+                set_schema_context(schema, "request", request)
                 deserialized = schema.load(value)
                 return deserialized
```
Shrey-N:
```suggestion  
          def _deserialize(self, value, attr, data, **kwargs):
                token = cornice_request.set(request)
                try:
                    set_schema_context(schema, "request", request)
                    deserialized = schema.load(value)
                    return deserialized
                finally:
                    cornice_request.reset(token)
```
Hiya @tomascorrea,
I believe calling `cornice_request.set(request)` without ever calling `.reset()` can cause a serious memory leak in WSGI servers that reuse worker threads across multiple requests. The `request` object will be kept alive indefinitely. 
You should wrap the schema loading in a `try...finally` block to safely reset the context token:

jenstroeger:
@tomascorrea can you please fix the lint?

